PR #10, opened 2019-11-22
Add template`s routes

oleh-momot:
It needs to change the destination into `feature/routes` after `feature/app-routes` will be merged


registry/server/templates/services/prepareTemplate.ts
```diff
@@ -0,0 +1,11 @@
+import _ from 'lodash/fp';
+
+import Template from '../interfaces';
+
+const convertTemplateContent = _.curry(((from: BufferEncoding, to: BufferEncoding, template: Template): Template => ({
+    ...template,
+    content: Buffer.from(template.content, from).toString(to),
+})));
+
+export const prepareTemplateToInsert = convertTemplateContent('utf-8', 'hex');
```
StyleT: We don't need `hex` here.

oleh-momot: Done

PR #15, opened 2019-11-27
chore: cli runner appAssetsDiscovery

nightnei:


registry/server/runnerAppAssetsDiscovery.ts
```diff
@@ -0,0 +1,5 @@
+const tmpInterval = Number(process.env.interval);
```
StyleT: INTERVAL

PR #23, opened 2019-12-11
fix: registry API bugs and improve responses info

nightnei:


registry/server/common/services/validateRequest.ts
```diff
@@ -26,8 +26,9 @@ const validateRequestFactory = (validationConfig: ValidationConfig[]) => async (
             async ({schema, selector}) => schema.validateAsync(selector(req)),
             validationConfig
         ));
-    } catch (error) {
-        res.status(422).send(preProcessErrorResponse(error));
+    } catch (e) {
+        res.status(422).send(preProcessErrorResponse(e));
+        throw e;
```
StyleT: Remove throw from here. Move to the caller side as discussed

registry/server/appRoutes/routes/updateAppRoute.ts
```diff
@@ -38,14 +38,25 @@ const validateRequestBeforeUpdateAppRoute = validateRequestFactory([
...
+        return;
+    }
 
     const {
         slots: appRouteSlots,
         ...appRoute
     } = req.body;
     const appRouteId = req.params.id;
 
+    const countToUpdate = await db('routes').where('id', appRouteId);
```
oleh-momot: As I see, we have this checking inside of the transaction
Please look at line 63

registry/server/common/services/validateRequest.ts
```diff
@@ -26,8 +26,8 @@ const validateRequestFactory = (validationConfig: ValidationConfig[]) => async (
             async ({schema, selector}) => schema.validateAsync(selector(req)),
             validationConfig
         ));
-    } catch (error) {
-        res.status(422).send(preProcessErrorResponse(error));
+    } catch (e) {
```
oleh-momot: I have one notice here
I want you to pay attention to consistent because of as I see you wrote `e` in the second file you wrote `err` also we already have `error`
So I suggest to follow a naming convention in this case

registry/server/templates/routes/updateTemplate.ts
```diff
@@ -30,11 +30,22 @@ const validateRequestBeforeUpdateTemplate = validateRequestFactory([
...
+        await validateRequestBeforeUpdateTemplate(req, res);
+    } catch(err) {
+        res.status(422).send(err);
+        return;
+    }
 
     const template = req.body;
     const templateName = req.params.name;
 
+    const countToUpdate = await db('templates').where({ name: templateName });
```
oleh-momot: I guess, this name of the variables which you wrote does not describe what inside of this variable
Because I guess here is `templates` well, they are entities, so `count` in this case does not describe what exactly we have there
I suggest to pay attention to the names of these variables in each of file

PR #24, opened 2019-12-11
tests: registry API

nightnei:


registry/test/apps.js
```diff
@@ -0,0 +1,388 @@
...
+
+        it('should not create record without a required field: spaBundle', async () => {
+            const response = await request.post(example.url)
+            .send(_.omit(example.correct, 'spaBundle'))
+            .expect(422, '"spaBundle" is required');
+
+            expect(response.body).deep.equal({});
+        });
+
+        it('should not create record with incorrect type of field: name', async () => {
```
oleh-momot: Well, you wrote good test cases
Actually, I have one notice, it would be good if you have only one expectation each of test case
It is antipattern to have bigger than one expect in one test case
I suggest to separate test case into several
Please check all of your test cases

registry/test/apps.js
```diff
@@ -0,0 +1,388 @@
...
+
+describe(`Tests ${example.url}`, () => {
+    before('should work simple "create" and "delete"', async () => {
+        await request.post(example.url).send(example.correct).expect(200)
+        await request.get(example.url + example.encodedName).expect(200);
+        await request.delete(example.url + example.encodedName).expect(204);
+        await request.get(example.url + example.encodedName).expect(404);
+    });
+    afterEach(example.delete);
+    describe('Create', () => {
```
oleh-momot: In my opinion, this name of this test suite is not a good
The same is for all test suites and test cases
I guess it should describe it from a user perspective
So, for example, `when a user is creating a new app` so after a test case describes what should be in special case `should create something` or `should get a validation error` etc.

registry/test/apps.js
```diff
@@ -0,0 +1,388 @@
...
+
+            expect(response.body).deep.equal({});
+
+            response = await request.get(example.url + example.incorrect.name)
+            .expect(404, 'Not found');
+
+            expect(response.body).deep.equal({});
+        });
+
+        it('should not create record with incorrect type of field: spaBundle', async () => {
```
oleh-momot: I guess we can create a test suite with the checking of validation
So, for example, you can have `when a user sends incorrect data` is a name of the test suite
Well a test case can have a name `should send a response with a validation error` or different
So the second test suite `when a user sends correct data` with a test case for example `should create an app` etc.

registry/test/appRoutes.ts
```diff
@@ -0,0 +1,322 @@
...
+
+describe(`Tests ${example.url}`, () => {
+    before('should work simple "create" and "delete"', async () => {
+        await example.create();
+        await request.get(example.url + example.correctId).expect(200);
+        await request.delete(example.url + example.correctId).expect(204);
+        await request.get(example.url + example.correctId).expect(404);
+        example.correctId = null;
+    });
+    afterEach(example.delete);
```
StyleT: Don't do auto-delete

registry/test/appRoutes.ts
```diff
@@ -0,0 +1,322 @@
...
+        if (example.correctId) throw new Error('Record has already been created');
+        const response = await request.post(example.url).send(example.correct);
+        example.correctId = response.body.id;
+        return response;
+    },
+    delete: async () => {
+        await request.delete(example.url + example.correctId);
+        example.correctId = null;
+    },
+    correctId: null,
```
StyleT: Remove this global var

registry/test/appRoutes.ts
```diff
@@ -0,0 +1,322 @@
...
+                appName: '@portal/ncTestRouteSlotNavbarUpdated',
+            },
+            ncTestRouteSlotBody: {
+                appName: '@portal/ncTestRouteSlotSystemUpdated',
+                props: { ncTestProp: 2 }
+            }
+        }
+    }),
+    generateExpectedResponseRoute: (id: Number | null) => ({ id, ..._.omitBy(example.correct, _.isNil) }),
+    create: async () => {
```
StyleT: rm this function

registry/test/appRoutes.ts
```diff
@@ -0,0 +1,211 @@
...
+};
+
+describe(`Tests ${example.url}`, () => {
+    before('should work simple "create" and "delete"', async () => {
+        const response = await request.post(example.url).send(example.correct);
+        await request.get(example.url + response.body.id).expect(200);
+        await request.delete(example.url + response.body.id).expect(204);
+        await request.get(example.url + response.body.id).expect(404);
+    });
+    describe('Create', () => {
```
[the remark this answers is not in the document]
nightnei: I check it in `should not create record without required fields('orderPos', 'route', 'slots')`

PR #25, opened 2019-12-12
Revert removing of SystemJS CSS extra plugin

oleh-momot:
**Important**! Also, need to re-check it when you are going to review it


oleh-momot:
SystemJS CSS extra plugin was reverted

PR #27, opened 2019-12-16
Filter slots to register unique apps

oleh-momot:


ilc/client.js
```diff
@@ -28,42 +28,70 @@ const router = new Router(registryConf);
 let currentPath = router.match(window.location.pathname + window.location.search);
 let prevPath = currentPath;
 
-registryConf.routes.map((route) => route.slots).forEach((slots) => {
-    Object.keys(slots).forEach((slotName) => {
-        const appName = slots[slotName].appName;
+registryConf.routes
```
StyleT: Move this complex logic to separate function. Ex: `getAppsToRegister()`

oleh-momot: Moved

PR #30, opened 2019-12-17
Now ILC works at BSB

StyleT:


registry/config/default.ts
```diff
@@ -4,7 +4,7 @@ module.exports = {
         client: 'sqlite3',
         connection: {
             database: 'registry_db',
-            filename: './server/mydb.sqlite',
+            filename: './server/dbfiles/db.sqlite',
```
oleh-momot: Did you move this sqlite file to that folder?

StyleT: Yes, we need this to simplify integration with Docker. It has some issues with file mounting while for folders everything works fine. So instead of mounting `./server/mydb.sqlite` now can mount `./server/dbfiles/` folder.

PR #31, opened 2019-12-18
On-the-fly conversion of the AMD modules that use "define" into "ILC.define"

oleh-momot:
Remove wrappers in Webpack configs and move wrappers sense into system js plugin

### DIFF
--- a/ilc/systemjs/systemjs-transform-loader.js
+++ b/ilc/systemjs/systemjs-transform-loader.js
@@ -1,0 +1,34 @@
+(function (global) {
+    const systemJSPrototype = global.System.constructor.prototype;
+    const instantiate = systemJSPrototype.instantiate;
+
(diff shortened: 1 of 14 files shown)


ilc/systemjs/systemjs-transform-loader.js
```diff
@@ -0,0 +1,34 @@
+(function (global) {
```
StyleT: I think you don't need to reinvent https://github.com/systemjs/systemjs/blob/master/dist/extras/transform.js here

StyleT: just use it from SystemJS bundle

oleh-momot: We have issues with fetching of CSS bundles by this module which does not handle them and it wants to eval them
So that is why we created ourselves plugin

PR #42, opened 2020-01-13
Feature/mysql support & OKD deployment fixes

StyleT:


registry/server/migrations/20191016185102_apps.ts
```diff
@@ -6,10 +6,10 @@ export async function up(knex: Knex): Promise<any> {
         table.string('name', 50).notNullable().primary();
         table.string('spaBundle', 255).notNullable();
         table.string('cssBundle', 255).nullable();
-        table.json('dependencies').defaultTo('{}');
+        table.json('dependencies');
```
nightnei: is it fix for mysql? I think it can cause some errors.

registry/Dockerfile
```diff
@@ -6,7 +6,7 @@ ADD ./ /codebase
 WORKDIR /codebase
 
 RUN npm install --no-package-lock --no-progress --ignore-scripts
-RUN npm install sqlite3
+RUN npm install sqlite3 mysql
```
nightnei: is sqlite needed only for local development? if it's correct, we don't need to install it here, do we?

registry/server/routes/config.ts
```diff
@@ -8,7 +8,7 @@ router.get('/', async (req, res) => {
     const [apps, templates, routes] = await Promise.all([
         knex.select().from('apps'),
         knex.select().from('templates'),
-        knex.select('routes.id as routeId', '*')
```
oleh-momot: image
Does this code work correctly with this change?

oleh-momot: Discussed

PR #53, opened 2020-01-24
Logic to handle errors more strictly in case we're dealing with Bots

StyleT:
…M bots

### DIFF
--- a/ilc/server/tailor/seobots-guard-stream.js
+++ b/ilc/server/tailor/seobots-guard-stream.js
@@ -1,0 +1,113 @@
+const stream = require('stream');
+const BotDetector = require('device-detector-js/dist/parsers/bot');
+const botDetector = new BotDetector();
+
+/**
+ * The goals of this stream is to change behaviour of the Tailor when dealing with SEO/SM bots.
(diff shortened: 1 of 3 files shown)


ilc/server/tailor/seobots-guard-stream.js
```diff
@@ -0,0 +1,113 @@
...
+        this.#batch.push(chunk);
+
+        if (this.#fragmentsCounter === 0 && this.#fragmentFound) {
+            this.#flushBuffer();
+        }
+
+        done();
+    }
+
+    _flush(done) {
```
nightnei: looks like it's unused method

ilc/server/tailor/seobots-guard-stream.js
```diff
@@ -0,0 +1,113 @@
...
+    #erroredState = false;
+    #writeHeadArgs = null;
+
+    constructor(reqHeaders, res) {
+        super({objectMode: true, autoDestroy: true});
+        this.#isBot = botDetector.parse(reqHeaders['user-agent']) !== null;
+        this.#response = res;
+    }
+
+    _transform(chunk, encoding, done) {
```
nightnei: looks like it's unused method

PR #67, opened 2020-02-19
Add error handler as a prop to app slots

oleh-momot:


adapters/vuejs/src/single-spa-vue.js
```diff
@@ -57,6 +50,10 @@ function bootstrap(opts, mountedInstances, props) {
         fragmentName = props.name;
     }
 
+    if (props.errorHandler) {
```
StyleT: I think if props.errorHandler is undefined you need to throw an error

StyleT: And in any case this code will fail if here you will take else path ;)

ilc/client/errorHandler/fragmentErrorHandlerFactory.js
```diff
@@ -0,0 +1,66 @@
+import * as uuidv4 from 'uuid/v4';
+
+import noticeError from './noticeError';
+import registryService from '../registry/factory';
+
+function fragmentErrorHandlerFactory(registryConf, getCurrentPath) {
```
nightnei: remove the second factory and use 4 arguments in the first

PR #72, opened 2020-03-02
Feature/docs update

StyleT:


nightnei:
let's add info about fastify

PR #86, opened 2020-03-09
fix: transition from vue to react

nightnei:


adapters/vuejs/src/single-spa-vue.js
```diff
@@ -82,16 +82,10 @@ function mount(opts, mountedInstances, props) {
...
-            if (!spaContainer) {
-                const singleSpaContainer = document.createElement('div');
-                singleSpaContainer.className = 'single-spa-container';
-
-                mountPoint.appendChild(singleSpaContainer);
-
-                spaContainer = singleSpaContainer;
-            }
-            mountPoint = spaContainer;
+            const singleSpaContainer = document.createElement('div');
```
StyleT: This element may already exist after SSR

PR #88, opened 2020-03-10
fix: empty link header

nightnei:


registry/server/templates/services/renderTemplate.ts
```diff
@@ -109,9 +109,12 @@ async function fetchIncludes(includesAttributes: IncludesAttributes): Promise<Ar
                 timeout: +timeout,
             });
 
-            const stylesheets = selectStylesheets(link);
+            if (link) {
```
oleh-momot: Please cover this one with a test

registry/server/templates/services/renderTemplate.spec.ts
```diff
@@ -18,13 +18,13 @@ describe('renderTemplate', () => {
...
                                 This include has all necessary attributes
-                                and a specified link header which is a stylesheet
+                                and an empty link header which is a stylesheet
                             </div>
                         `,
                         headers: {
                             'X-Powered-By': 'JS',
                             'X-My-Awesome-Header': 'Awesome',
-                            'Link': 'https://my.awesome.server/my-awesome-stylesheet.css;rel=stylesheet;loveyou=3000',
+                            'Link': '',
```
StyleT: Add new request instead of modification of the existing one.
Also test that not only empty `Link` but also absence of `Link` works

registry/server/templates/services/renderTemplate.spec.ts
```diff
@@ -311,6 +311,54 @@ describe('renderTemplate', () => {
         `);
     });
 
+    it('should return a rendered template with has all necessary attributes and an empty link header ', async () => {
```
StyleT:
```suggestion
    it('should return a rendered template with has all necessary attributes but without link header ', async () => {
```

PR #96, opened 2020-03-18
fix: cached pages

nightnei:


ilc/client.js
```diff
@@ -144,10 +144,20 @@ function getCurrentBasePath() {
     return currentPath.basePath;
 }
 
+function getNormalizedLocation() {
+    // we should respect base tag for cached pages
+    const base = document.querySelector('base');
```
StyleT: Use it only during init & remove after first use while showing warning in console.

oleh-momot: We can specify where exactly we should find 'base' to speed up it
e.g. document.head.querySelector or document.body.querySelector

nightnei: 1) `base` tag should be located in `head`, but this option can cover both `head` and `body`.
2) additionally it useless optimization, first of all, because `base` tag should be in `head`, which is the first tag in document.

oleh-momot: As I understood correctly, it should be always in the head
So I don't see any point to search it in both sources

ilc/client.js
```diff
@@ -144,11 +145,36 @@ function getCurrentBasePath() {
     return currentPath.basePath;
 }
 
+
+let prevUrl;
+let currentUrl = window.location.pathname + window.location.search;
 window.addEventListener('single-spa:before-routing-event', () => {
+    // !!! TODO remove it after fix https://github.com/single-spa/single-spa/issues/484
```
StyleT:
```suggestion
    //TODO: remove it after fix https://github.com/single-spa/single-spa/issues/484
```

ilc/client.js
```diff
@@ -144,11 +145,36 @@ function getCurrentBasePath() {
...
+    // and add "base.remove();" after first time of using
+    prevUrl = currentUrl;
+    currentUrl = window.location.pathname + window.location.search;
+    if (currentUrl !== prevUrl) {
+        const base = document.querySelector('base');
```
StyleT: Search in head + I would prefer `getElementsByTagName` here

ilc/client.js
```diff
@@ -144,11 +145,36 @@ function getCurrentBasePath() {
...
+        if (base) {
+            base.remove();
+            console.warn('Base tag is removed!!!');
```
StyleT: ??? what is this?

StyleT:
```suggestion
            console.warn('ILC: <base> tag was used only for initial rendering & removed afterwards. Currently we\'re not respecting it fully. Pls open an issue if you need this functionality.');
```

ilc/client.js
```diff
@@ -144,11 +145,36 @@ function getCurrentBasePath() {
...
+        }
+    }
+
     prevPath = currentPath;
```
StyleT: Now `prevPath` will be `undefined` before 1st reroute from what I understand

ilc/common/router/ServerRouter.js
```diff
@@ -49,7 +48,7 @@ module.exports = class ServerRouter {
     #getRouter = async () => {
         const now = Math.floor(Date.now() / 1000);
 
-        if (this.#registryConf === null || this.#checkAfter < now) {
+        if (this.#checkAfter < now) {
```
StyleT:
```suggestion
        if (this.#router === null || this.#checkAfter < now) {
```

ilc/common/router/Router.js
```diff
@@ -55,7 +55,7 @@ module.exports = class Router {
         });
     }
 
-    __compiler(routes) {
+    #compiler = (routes) => {
```
StyleT: I think there was some issue with private class properties & babel in context of an old browsers. 
@oleh-momot do you remember it?

ilc/client.js
nightnei: I would prefer getElementsByTagName here
1) getElementsByTagName return array of all items and we will take the first one. querySelector is faster because it looks only for the first one.
2) querySelector return static array of NodeList unlike getElementsByTagName returns dynamic HTMLCollection.

Search in head 
1) search engines put head tags of the page to body. so it's better to search in head and body.

ilc/client.js
nightnei: As I understood correctly, it should be always in the head
-- no, it can be in body too.

ilc/common/router/Router.js
StyleT: Looks like private props will be complied onto a thing using WeakSet... Which is not supported in IE.
https://babeljs.io/docs/en/babel-plugin-proposal-private-methods
https://caniuse.com/#search=Weakset

ilc/client.js
```diff
@@ -1,7 +1,6 @@
 import * as singleSpa from 'single-spa';
-import deepmerge from 'deepmerge';
 
-import * as Router from './common/router/Router';
+import Router from './common/router/ClientRouter';
```
oleh-momot: As I understood, we use it only on the client-side
Please move it into the './client'

nightnei: we have router folder, so it's incorrect to put client router to separate folder

PR #101, opened 2020-03-20
Add compression to decrease client js bundle from 80 to 30 kB

oleh-momot:


ilc/server/serveStatic.js
```diff
@@ -17,10 +17,15 @@ module.exports = function (isProduction) {
             publicPath: '/',
             headers: {
                 "Access-Control-Allow-Origin": "*",
+                "Content-Encoding": "gzip",
```
nightnei: it's development, and now gzipping works by default on prod

oleh-momot: Yeah, I missed it
I'll add it
Thanks!

oleh-momot: Added

PR #109, opened 2020-03-26
Preload styles firstly and load scripts then

oleh-momot:
**It depends on https://github.com/StyleT/tailorx/pull/20/files**


ilc/server/tailor/configs-injector.js
```diff
@@ -28,27 +27,47 @@ module.exports = class ConfigsInjector {
             this.#getPolyfill(),
             this.#wrapWithAsyncScriptTag(this.#getClientjsUrl()),
```
oleh-momot: @StyleT Should we leave it?

PR #112, opened 2020-03-27
Add style refs of a rendered template to the response

oleh-momot:


ilc/server/tailor/configs-injector.js
```diff
@@ -54,10 +54,11 @@ module.exports = class ConfigsInjector {
     #getRouteStyleRefsToPreload = async (reqUrl) => {
         const registryConf = await this.#registry.getConfig();
         const route = await this.#router.getRouteInfo(reqUrl);
+        const baseTemplate = await this.#registry.getTemplate(route.template);
```
StyleT: Is there is a chance that here we'll get template that is different (new version) from the one we used inside `router.getTemplateInfo`?

ilc/server/tailor/configs-injector.js
```diff
@@ -6,19 +6,26 @@ module.exports = class ConfigsInjector {
     #registry;
     #router;
     #cdnUrl;
+    #temp;
 
     constructor(registry, router, cdnUrl = null) {
         this.#registry = registry;
         this.#router = router;
         this.#cdnUrl = cdnUrl;
+        this.#temp = {
```
StyleT: I will never work as we share instance of ConfigsInjector between requests

PR #120, opened 2020-04-06
Feature/registry admin poc

StyleT:
https://take.ms/wm6z4


install_all.js
```diff
@@ -4,6 +4,7 @@ const spawnSync = require('child_process').spawnSync;
 const commands = [
     { command: 'cd ./ilc/ && npm i', name: 'ilc' },
     { command: 'cd ./registry/ && npm i', name: 'registry' },
+    { command: 'cd ./registry/client && npm i', name: 'registry' },
```
nightnei, line 7: let's rename to eg `name: 'registry_admin'`

StyleT: Renamed to registry:ui to be consistent with `start_all.js`

StyleT: Renamed to registry:ui to be consistent with start_all.js

registry/tests/test.spec.ts
```diff
@@ -3,6 +3,6 @@ import { request, expect } from './common';
 describe('Tests /', () => {
     it('Hello world', async () => {
         const response = await request.get('/');
-        expect(response.text).to.be.equal('Hello! This is Micro Fragments registry service.');
+        expect(response.text).to.contain(' <title>React Admin</title>');
```
oleh-momot, line 6: I guess, space is not necessary here (' <title...')

StyleT: updated

PR #125, opened 2020-04-10
feat: lde via cookies

nightnei:


ilc/server/tailor/fetch-template.js
```diff
@@ -15,7 +17,19 @@ module.exports = (templatesPath, router, configsInjector, newrelic, registryServ
     request,
     parseTemplate
 ) => {
-    const registryConfig = await registryService.getConfig();
+    let registryConfig = await registryService.getConfig();
+
+    const { cookie } = request.headers;
+    try {
+        let overrideConfig = typeof cookie === 'string' && cookie.split('; ').find(n => n.startsWith('overrideConfig'));
```
StyleT, line 24: For extra safety:
```suggestion
        let overrideConfig = typeof cookie === 'string' && cookie.split(';').find(n => n.trim().startsWith('ILC-overrideConfig'));
```

ilc/server/tailor/fetch-template.js
```diff
@@ -15,7 +17,19 @@ module.exports = (templatesPath, router, configsInjector, newrelic, registryServ
     request,
     parseTemplate
 ) => {
-    const registryConfig = await registryService.getConfig();
+    let registryConfig = await registryService.getConfig();
+
+    const { cookie } = request.headers;
```
StyleT, line 22: Let's expect cookie value to be URL encoded. To mitigate risk of having issues because of the special characters inside JSON.

ilc/server/tailor/fetch-template.js
```diff
@@ -15,7 +17,19 @@ module.exports = (templatesPath, router, configsInjector, newrelic, registryServ
...
+    try {
+        let overrideConfig = typeof cookie === 'string' && cookie.split('; ').find(n => n.startsWith('overrideConfig'));
+        if (overrideConfig) {
+            overrideConfig = JSON.parse(overrideConfig.replace(/^overrideConfig=/, ''));
+            mergeConfigs(registryConfig.data, overrideConfig);
+        }
+    } catch (e) {
+        console.error('Can\'t handle overrideConfig from cookies', e);
```
StyleT, line 30: Just ignore values that you failed to decode. No console.error is necessary.

ilc/server/tailor/merge-configs.js
```diff
@@ -0,0 +1,15 @@
+const deepmerge = require('deepmerge');
+
+module.exports = (original, override) => {
```
StyleT, line 3: You need to do a deep clone of the original object before it's modification. As same object may be shared between requests.

ilc/server/tailor/merge-configs.js
```diff
@@ -0,0 +1,15 @@
...
+    original.apps = deepmerge(original.apps, override.apps);
```
StyleT, line 4: What if override.apps does not exists?

ilc/server/tailor/merge-configs.js
nightnei: I planed to do it, but I don't see any case to reproduce it, so decided to speed up with don't do cloning. but you are right, it's better to do it, it's more safe

PR #144, opened 2020-04-21
chore: docs of trusted origins and practical example

nightnei:


docs/develop_at_production.md
```diff
@@ -1,5 +1,33 @@
 # Develop right at production
 
+
+## Practical example of using existed MS on our [demo site](http://demo.microfrontends.online/news/)
```
StyleT, line 4: I think this guide is too vague... I would expect it to have list of commands/steps I can copy/paste and get result.
Also I don't see any mention here that if you use NAT - you need ngrok

nightnei: > Also I don't see any mention here that if you use NAT - you need ngrok

I added it in comment https://github.com/namecheap/ilc/pull/144/files#diff-b6dabd60a2870f92100d7ccdcdae3509R24

> I think this guide is too vague... I would expect it to have list of commands/steps I can copy/paste and get result.
You mean I should duplicate even commands from guide of setupping ilc-demo-apps?

PR #152, opened 2020-04-29
Run tests when Docker tries to build ILC image

oleh-momot:


ilc/Dockerfile
```diff
@@ -1,14 +1,32 @@
...
-ADD ./ /codebase
-WORKDIR /codebase
+ENV CHROME_BIN=/usr/bin/chromium-browser
+ENV CHROME_DRIVER=/usr/bin/chromedriver
+
+ADD ./ /temporary
+WORKDIR /temporary
 
 RUN npm install --no-package-lock --no-progress
+RUN npm run test
```
StyleT, line 12: I don't really like the approach of running tests within docker build process as it makes impossible to extract test artefacts & attach them as CI run results

StyleT: Look at the https://docs.docker.com/engine/reference/commandline/build/#specifying-target-build-stage---target

ilc/Dockerfile
```diff
@@ -1,14 +1,32 @@
...
+ADD ./ /codebase
+WORKDIR /codebase
+
+COPY --from=build /temporary/package.json ./package.json
+COPY --from=build /temporary/server ./server
+COPY --from=build /temporary/common ./common
+COPY --from=build /temporary/public ./public
+COPY --from=build /temporary/config ./config
+
+RUN npm install --no-package-lock --no-progress --production
```
StyleT, line 30: Think how we can ensure that we have all necessary deps installed and no one accidentally marked a "real" dependency as dev one.

ilc/package.json
```diff
@@ -5,11 +5,11 @@
...
     "test": "cross-env NODE_ENV=test mocha",
-    "test:watch": "cross-env NODE_ENV=test mocha --watch",
+    "test:watch": "npm run test -- --watch",
     "test:coverage": "cross-env NODE_ENV=test nyc mocha",
-    "test:client": "cross-env NODE_ENV=test karma start --single-run",
-    "test:client:watch": "cross-env NODE_ENV=test karma start --no-single-run",
-    "start": "node --max-http-header-size 30000 server/app.js",
+    "test:client": "npm run build:systemjs && cross-env NODE_ENV=test karma start",
+    "test:client:watch": "npm run test:client -- --no-single-run",
+    "start": "NODE_ENV=production node --max-http-header-size 30000 server/app.js",
```
StyleT, line 12: This is not ok as it prevents k8s to override NODE_ENV

oleh-momot: We have the issue with `config` package
This package tries to get development config instead of default when NODE_ENV is not provided even if don't provide NODE_ENV=development

StyleT: Yes, I know it, but nevertheless we can't keep it here

ilc/Dockerfile
vetal2409: Agree, In the multistage building it is impossible to fetch artifacts from the not final stage. --target can solve the problem but then it's required to solve the "copy" problem. By "copy" problem I mean either to run a container and copy data OR use --out option for the build command but it will copy all the filesystem to the host.

Probably, in this case, it's better to split into two docker files. The first one with chrome, we build, create a container, and run the tests with the mounted volume (to get the artifacts on the host). 
And the second one is clean Dockerfile without test dependencies (which is the main goal as I understand)

ilc/Dockerfile
oleh-momot: I tried several packages that can help but unfortunately, no one suits
So I decided to remove these changes and restore previous.

.github/workflows/ci.yml
```diff
@@ -21,11 +21,32 @@ jobs:
...
         cd ./ilc/
-        docker build . --tag ilc:tmp
+        DOCKER_BUILDKIT=1 docker build . --tag ilc:tmp
 
-    - name: Run tests
+    - name: Run sever side tests
       run: docker run ilc:tmp npm run test
 
+    - name: Run server side tests coverage
+      run: docker run --volume /artifacts/ilc/.test_output/server:/temporary ilc:tmp bash -c "npm run test:coverage && mv .nyc_output/* /temporary"
```
StyleT, line 30: Just curious why do you need to use such a long path `/artifacts/ilc/.test_output/server`? Can't it be just `/artifacts/server`?

oleh-momot: I decided to specify it in more detail because we may also have artifacts for registry

PR #153, opened 2020-05-04
Cover handle page transaction with tests

oleh-momot:
Need to change dest to master after `feature/run-tests-on-ci` will be fixed and merged


ilc/client/handlePageTransaction.js
```diff
@@ -78,19 +78,31 @@ const renderFakeSlot = slotName => {
...
 
-    if (willBe === 'rendered') {
-        addContentListener(slotName);
-    } else if (willBe === 'removed') {
-        renderFakeSlot(slotName);
-    } else if (willBe === 'rerendered') {
-        renderFakeSlot(slotName);
-        addContentListener(slotName);
+    switch (slotWillBe) {
+        case slotCanBe.rendered: {
```
StyleT, line 91: I'm pretty sure that you don't need `{ ... }` brackets for case statements. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch

ilc/client/handlePageTransaction.spec.js
```diff
@@ -0,0 +1,258 @@
+import chai from 'chai';
+import sinon from 'sinon';
+import html from 'nanohtml';
+
+import handlePageTransaction, {
+    slotCanBe,
+} from './handlePageTransaction';
+
+describe.only('handle page transaction', () => {
```
StyleT: rm only

ilc/client/handlePageTransaction.spec.js
```diff
@@ -0,0 +1,255 @@
...
+            <div id="${applications.navbar.id}" class="${applications.navbar.class}">
+                Hello! I am Navbar SPA application
+            </div>`;
+
+        applications.body.ref = html`
+            <div id="${applications.body.id}" class="${applications.body.class}">
+                Hello! I am Body SPA application
+            </div>`;
+
+        slots.ref = html`
```
StyleT, line 66: Could you define this property at the same time you declare `slots` variable? Now this code isn't really obvious.,

ilc/client/handlePageTransaction.spec.js
```diff
@@ -0,0 +1,255 @@
...
+
+        clock = sinon.useFakeTimers();
+    });
+
+    afterEach(() => {
+        slots.removeSlots();
+        clock.restore();
+    });
+
+    it('should do nothing when a slot name or a slot action are not provided', async () => {
```
StyleT, line 85: Shouldn't we raise an error here instead of doing nothing?

ilc/client/handlePageTransaction.spec.js
```diff
@@ -0,0 +1,255 @@
...
+        await clock.runAllAsync();
+
+        chai.expect(spinner.getRef()).to.be.null;
+        chai.expect(slots.ref.innerHTML).to.be.equal(
+            `<div id="${slots.navbar.id}"></div>` +
+            `<div id="${slots.body.id}"></div>`
+        );
+    });
+
+    it('should do nothing when a slot action does not match any possible option to handle', async () => {
```
StyleT, line 97: I would make it accept null (default parameter value) or valid action only. It makes more sense from the code readability perspective.

ilc/client.js
```diff
@@ -3,7 +3,7 @@ import * as singleSpa from 'single-spa';
 import Router from './common/router/ClientRouter';
 import setupErrorHandlers from './client/errorHandler/setupErrorHandlers';
 import {fragmentErrorHandlerFactory, crashIlc} from './client/errorHandler/fragmentErrorHandlerFactory';
-import handlePageTransaction from './client/handlePageTransaction';
+import handlePageTransaction, {slotCanBe} from './client/handlePageTransaction';
```
StyleT, line 6: `slotCanBe` -> `slotWillBe`

StyleT: As you already use `willBe` below

ilc/client/handlePageTransaction.js
oleh-momot: Yes, it is not necessary
It is necessary only when you have a let or const variables that you may declare inside of brackets there

ilc/client/handlePageTransaction.spec.js
oleh-momot: It is a good point, I thought about it and wanted to discuss it
I am going to add throwing an error

ilc/client/handlePageTransaction.spec.js
oleh-momot: Yeah, I agree
I am going to change it

ilc/client/handlePageTransaction.spec.js
```diff
@@ -0,0 +1,278 @@
...
+
+        chai.expect(caughtError).to.be.an.instanceof(Error);
+        chai.expect(caughtError.message).to.be.eql('A slot name was not provided!');
+    });
+
+    it('should throw an error when a slot action does not match any possible option to handle', async () => {
+        let caughtError;
+
+        try {
+            handlePageTransaction(slots.body.id, 'undefined');
```
StyleT, line 111: `chai.expect(() => handlePageTransaction(slots.body.id, 'undefined')).to.throw()`
https://www.chaijs.com/api/bdd/#method_throw
Instead of try/catch

PR #154, opened 2020-05-06
Cover async boot up with karma tests

oleh-momot:
Need to change dest to `master` when `cover/handle-page-transaction` is going to be merged


ilc/client/asyncBootUp.spec.js
```diff
@@ -0,0 +1,147 @@
...
+        clock = sinon.useFakeTimers();
+    });
+
+    afterEach(() => {
+        clock.restore();
+        document.body.removeChild(slots.ref);
+        overrideImportMap.resetHistory();
+    });
+
+    it('should wait for slots before single-spa`s routing-event, mark apps as ready if they exist and return override configs then', async () => {
```
StyleT, line 95: 1. I don't think that it's acceptable to have a single test to cover all the logic within `asyncBootup.js`. This particular test is too generic and it's hard to understand it.
2. You actually haven't tested quite a lot of the use cases 
image

PR #156, opened 2020-05-06
Cover is active factory with karma tests

oleh-momot:
Need to change dest to `master` after `cover/async-boot-up` is going to be merged

### DIFF
--- a/ilc/client/isActiveFactory.js
+++ b/ilc/client/isActiveFactory.js
@@ -1,11 +1,7 @@
-export default function isActiveFactory({
-    singleSpa,
-    router,
-    handlePageTransaction,
-    slotWillBe,
-    appName,
-    slotName,
-}) {
+import {triggerAppChange} from 'single-spa';
+import handlePageTransaction, {slotWillBe} from './handlePageTransaction';
+
+export const createFactory = (triggerAppChange, handlePageTransaction, slotWillBe) => (router, appName, slotName) => {
     let reload = false;
 
     return () => {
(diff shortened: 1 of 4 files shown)


ilc/client/isActiveFactory.js
```diff
@@ -0,0 +1,52 @@
+export default function isActiveFactory({
+    singleSpa,
```
StyleT, line 2: I think that you can use imports to handle:
- `singleSpa`
- `handlePageTransaction`
- `slotWillBe`

ilc/client/isActiveFactory.spec.js
```diff
@@ -0,0 +1,168 @@
...
+
+        chai.expect(isActive()).to.be.true;
+
+        dispatchSingleSpaAppChangeEvent();
+
+        await clock.runAllAsync();
+
+        chai.expect(singleSpa.triggerAppChange.called).to.be.false;
+        chai.expect(handlePageTransaction.calledOnceWithExactly(slotName, slotWillBe.rendered)).to.be.true;
+        chai.expect(router.getPrevRouteProps.called).to.be.false;
```
StyleT, line 98: I don't think that you need to check that:
- `chai.expect(router.getPrevRouteProps.called).to.be.false;`
- `chai.expect(router.getCurrentRouteProps.called).to.be.false;`

These are read operations and have no direct connection to business logic.

StyleT: Same below

PR #160, opened 2020-05-14
Cover system js override import map with karma tests

oleh-momot:
Change the dest to `master` when `cover/system-js-name-resolver` is going to be merged


ilc/systemjs/systemjs-override-importmap.spec.js
```diff
@@ -0,0 +1,45 @@
+import chai from 'chai';
+
+describe.only('systemjs override import map', () => {
```
StyleT, line 3: rm only

oleh-momot: Removed

PR #167, opened 2020-05-21
Cover error handler with tests

oleh-momot:


ilc/server/tailor/error-handler.spec.js
```diff
@@ -0,0 +1,149 @@
+const chai = require('chai');
+const sinon = require('sinon');
+const proxyquire = require('proxyquire').noPreserveCache();
```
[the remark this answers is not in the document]
oleh-momot: I removed `proxyquire` and changed it to DI.

ilc/server/tailor/error-handler.js
```diff
@@ -37,15 +37,18 @@ function handleFragmentWarn(req, fragmentAttrs, err) {
     noticeError(new errors.FragmentWarn(errOpts));
 }
 
-/**
- * Setup error handlers for Tailor
- * @param {Tailor} tailor
- */
-module.exports = function setup(tailor) {
+const preSetupErrorHandlers = (errors, noticeError, errorHandler) => (tailor) => {
```
StyleT, line 40: Pls revert all the `function in function` things you've added here. Same thing can be done in a much simpler way.
```
module.exports = function setup(tailor, errorHandler, noticeError) {
```

And in `tailor/factory.js` you will have:
```

const appErrorHandler = require('../errorHandler');
const noticeError = require('../errorHandler/noticeError');
const errorHandlerSetup = require('./error-handler');

errorHandlerSetup(tailor, appErrorHandler, noticeError);
```

oleh-momot: I changed setup function, reverted error handlers, and moved them into setup function to use dependencies from there.

PR #168, opened 2020-05-21
Cover notice error with tests

oleh-momot:


ilc/server/errorHandler/index.spec.js
```diff
@@ -0,0 +1,126 @@
+const chai = require('chai');
```
StyleT, line 1: I don't think that we need unit tests for this... Integration tests should be used instead. Now test is too fragile

StyleT: Let's discuss this...

oleh-momot: I wrote integration tests
Need to re-review

PR #171, opened 2020-05-27
Cover configs injector with tests

oleh-momot:


ilc/server/tailor/configs-injector.spec.js
```diff
@@ -0,0 +1,429 @@
...
+                                    cssBundle: registryConfig.apps.secondApp.cssBundle,
+                                    dependencies: {
+                                        ...registryConfig.apps.secondApp.dependencies,
+                                    },
+                                    props: {
+                                        ...registryConfig.apps.secondApp.props,
+                                    },
+                                    kind: registryConfig.apps.secondApp.kind,
+                                },
+                                thirdApp: {
```
StyleT, line 373: Couldn't you just write `thirdApp: registryConfig.apps.thirdApp`?

StyleT: Maybe use _.pick() if you need only some props

oleh-momot: I changed it, so I created a function that should return SPA config. _.pick is used to get props of apps there.

PR #172, opened 2020-06-03
Cover routers with tests

oleh-momot:


ilc/client.js
```diff
@@ -17,7 +17,7 @@ if (System === undefined) {
 }
 
 const registryConf = initSpaConfig();
-const router = new Router(registryConf);
+const router = new Router(registryConf, singleSpa.navigateToUrl);
```
nightnei, line 20: navigateToUrl is global var. why should we throw it inside router?

ilc/common/router/Router.js
```diff
@@ -64,17 +70,22 @@ module.exports = class Router {
...
-                v.routeExp = new RegExp(`^(${basePath})/?.*`);
+                routeExp = new RegExp(`^(${basePath})/?.*`);
             } else {
-                v.routeExp = new RegExp(`^(${route})$`);
+                routeExp = new RegExp(`^(${route})$`);
             }
 
-            return v;
+            return {
+                ...v,
```
nightnei, line 86: why should we create new object?

ilc/server/tailor/server-router.js
```diff
@@ -59,7 +60,7 @@ module.exports = class ServerRouter {
             url.searchParams.append('routerProps', Buffer.from(JSON.stringify(reqProps)).toString('base64'));
 
             if (slotData.props !== undefined || appInfo.props !== undefined) {
-                const appProps =  _.merge({}, appInfo.props, slotData.props);
+                const appProps = _.merge({}, appInfo.props, slotData.props);
```
nightnei, line 63: suggestion - let's use deepmerge everywhere to have codebase more consistency

ilc/common/router/Router.js
oleh-momot: Need to avoid any mutation (side effects).
This logic changes slots from the registry config.

ilc/client.js
oleh-momot: This is not a global var. It is part of single spa. So, we need to test the router with DI first of all. If you have any solution on how to test it without DI we can discuss it.

ilc/server/tailor/server-router.js
oleh-momot: It is a good point. I guess it is not a part of this PR. I am going to change it in another PR and find other places to reach consistency.

PR #189, opened 2020-08-03
Feature/better 404 page handling

StyleT:


e2e/spec/404.spec.e2e.ts
```diff
@@ -1,6 +1,83 @@
...
+    I.waitInUrl(newsPage.newsUrl, 10);
+    I.waitForElement(newsPage.goToNonExistingRouteOverride, 10);
+    I.click(newsPage.goToNonExistingRouteOverride);
+    I.waitForText('404 not found component', 10, 'body > div#body');
+});
+//endregion 404 page for non-existing News app route
+
+//region 404 page for non-existing News resource
+Scenario('Renders (SSR) global 404 page for non-existing News resource', (I) => {
+    I.amOnPage('/news/article/abc-news-au34');
```
nightnei, line 53: I propose:
1) to move all urls to file injectors
2) and inside those files create 2 objects: urls and selectors

I think it's more obviously.

StyleT: Did some refactoring with news urls, pls take a look. 
I think we need a separate PR for more refactoring around this topic.

ilc/client/ClientRouter.js
```diff
@@ -48,48 +51,72 @@ export default class ClientRouter {
...
+
         document.removeEventListener('click', this.#onClickLink);
     }
 
     #onSingleSpaRoutingEvents = () => {
         this.#prevRoute = this.#currentRoute;
 
+        const newUrl = this.#getCurrUrl();
+        if (this.#forceSpecialRoute !== null && this.#forceSpecialRoute.url === newUrl) {
+            this.#currentRoute = this.#router.matchSpecial(newUrl, this.#forceSpecialRoute.id);
```
nightnei, line 110: Can we reset "#forceSpecialRoute" immediately after using it here?
Why we should do it in separate `else if`?

ilc/server/tailor/configs-injector.js
```diff
@@ -149,7 +150,16 @@ module.exports = class ConfigsInjector {
         const apps = _.mapValues(registryConfig.apps, v => _.pick(v, ['spaBundle', 'cssBundle', 'dependencies', 'props', 'kind']));
         const spaConfig = JSON.stringify(_.omit({...registryConfig, apps}, ['templates']));
 
-        return `<script type="spa-config">${spaConfig}</script>`;
+        return `<script type="ilc-config">${spaConfig}</script>`;
+    };
+
+    #getIlcState = (request) => {
+        const state = request.ilcState || {};
```
oleh-momot, line 157: I don't understand when we would not have `ilcState`
Because I see `ilcState` is always expected here is ilc/server/tailor/error-handler.js and it would be added to the request here is ilc/server/app.js

e2e/spec/404.spec.e2e.ts
```diff
@@ -1,6 +1,83 @@
...
+    //After 404 page ILC continues normal operation
+    I.wait(5); //Hack to fix issue with the Vue Router
+    I.click(newsPage.linkWithUrl(newsPage.url.main));
+    I.waitForElement(newsPage.newsSources, 10);
+    I.see('Pick a news source', newsPage.bannerHeadline);
+});
+
+Scenario('Renders (SSR) overridden 404 page for non-existing News resource', (I, newsPage: newsPage) => {
+    I.amOnPage(newsPage.url.nonExistingResourceWithOverride);
+    I.waitForText('404 not found component', 1000, 'body > div#body');
```
oleh-momot, line 73: Do we exactly need to wait here for 1000 seconds?

e2e/spec/404.spec.e2e.ts
```diff
@@ -1,6 +1,83 @@
...
+    I.amOnPage(peoplePage.peopleUrl);
+    I.waitForElement(notFoundPageLink, 30);
+    I.click(notFoundPageLink);
+    I.waitForText('404 not found', 10, 'body > div#body');
+});
+//endregion 404 page for non-existing ILC route
+
+//region 404 page for non-existing News app route
+Scenario('Renders (SSR) global 404 page for non-existing News app route', (I) => {
+    I.amOnPage('/news/nonExisting');
```
oleh-momot, line 20: I see you added URLs here is `e2e/spec/pages/news.ts`
So you can use `newsPage.url.nonExistingRoute` here

e2e/spec/404.spec.e2e.ts
StyleT: 👍

e2e/spec/404.spec.e2e.ts
StyleT: 👍

PR #200, opened 2020-09-21
Feature/i18n/basic ssr implementation

StyleT:


ilc/server/i18n.js
```diff
@@ -0,0 +1,80 @@
+const _ = require('lodash');
+const cookie = require('cookie');
+
+const DEFAULT_LOCALE = 'en-US';
```
oleh-momot, line 4: Perhaps, it would be nice to have this and the code below in the config
What do you think about it?

ilc/server/tailor/factory.js
```diff
@@ -7,15 +7,14 @@ const logger = require('../logger');
...
-const registryService = require('../registry/factory');
 const filterHeaders = require('./filter-headers');
 const errorHandlingService = require('../errorHandler/factory');
 const errorHandlerSetup = require('./error-handler');
 const fragmentHooks = require('./fragment-hooks');
 const ConfigsInjector = require('./configs-injector');
 const processFragmentResponse = require('./process-fragment-response');
 
-module.exports = function (cdnUrl, nrCustomClientJsWrapper = null) {
+module.exports = function (cdnUrl, nrCustomClientJsWrapper = null, registryService) {
```
oleh-momot, line 17: I guess `registryService` should be the first param because `nrCustomClientJsWrapper` is `null` by default

nightnei: moreover, imho it's better to use object(e.g `{cdnUrl, nrCustomClientJsWrapper = null, registryService}`) here, since we have too many params, and some of them are optional

nightnei (approved):

ilc/server/tailor/factory.js
nightnei: moreover, imho it's better to use object(e.g {cdnUrl, nrCustomClientJsWrapper = null, registryService}) here, since we have too many params, and some of them are optional

StyleT: Changed args order. We can use obj syntax later IMO

oleh-momot (approved):

PR #203, opened 2020-09-23
feat: upgrade to the single-spa 5.6.0

StyleT:


ilc/client/isActiveFactory.js
```diff
@@ -1,4 +1,4 @@
-import {triggerAppChange} from 'single-spa';
+import {triggerAppChange} from './navigationEvents';
```
oleh-momot, line 1: Do we have here events setting up twice?
Look at `ilc/client` line number 1

ilc/client.js
```diff
@@ -1,3 +1,4 @@
+import './client/navigationEvents';
```
oleh-momot, line 1: It is a little bit tricky
I think that would be easy just to import `setupEvents` here

ilc/client/isActiveFactory.js
StyleT: Nope, due to the cache of the imports. Files are executed only once

ilc/client.js
StyleT: The idea is to have single entry point to this "module"

oleh-momot (approved):

nightnei (approved):

PR #210, opened 2020-10-07
fix: show spinner till content isnt visible

nightnei:


ilc/client/handlePageTransaction.js
```diff
@@ -51,13 +51,30 @@ const addContentListener = slotName => {
...
+
+        // if we have rendered MS to DOM but temporary hide it for some reason, e.g. to fetch data
+        if (!status.isAnyChildVisible) {
+            status.isAnyChildVisible = Array.from(targetNode.children).some(node => node.style.display !== 'none');
+        }
 
-        const hasText = !!targetNode.innerText.trim().length
-        const hasOpticNodes = !!targetNode.querySelector(':not(div):not(span)');
-        if (!hasText && !hasOpticNodes) return;
+        if (Object.values(status).some(n => !n)) return;
```
StyleT, line 77: Use if with brackets pls

StyleT (approved):

ilc/client/handlePageTransaction.js
oleh-momot, line 77: I found myself thinking that we probably do not have any polyfill for Object.values, so this would not work in IE11
So I suggest to test it in IE11 and change this by using supported methods or add polyfill for that if we still support IE11

ilc/client/handlePageTransaction.js
```diff
@@ -51,13 +51,30 @@ const addContentListener = slotName => {
...
+        // if we have rendered MS to DOM but meaningful content isn't rendered, e.g. due to essential data preload
+        if (!status.hasTextOrOpticNodes) {
+            const hasText = !!targetNode.innerText.trim().length
+            const hasOpticNodes = !!targetNode.querySelector(':not(div):not(span)');
+            status.hasTextOrOpticNodes = hasText || hasOpticNodes;
+        }
+
+        // if we have rendered MS to DOM but temporary hide it for some reason, e.g. to fetch data
+        if (!status.isAnyChildVisible) {
+            status.isAnyChildVisible = Array.from(targetNode.children).some(node => node.style.display !== 'none');
```
oleh-momot, line 74: IE11 does not support Array.from

oleh-momot (approved):

PR #215, opened 2020-10-13
Feature/i18n/client side api

StyleT:


ilc/client/isActiveFactory.spec.js
```diff
@@ -137,15 +124,13 @@ describe('is active factory', () => {
...
-
         chai.expect(isActive()).to.be.true;
 
         dispatchSingleSpaAppChangeEvent();
 
         await clock.runAllAsync();
 
         chai.expect(triggerAppChange.calledOnce).to.be.true;
-        chai.expect(handlePageTransaction.calledTwice).to.be.true;
+        chai.expect(handlePageTransaction.calledThrice).to.be.true;
```
oleh-momot, line 134: If it should call thrice, I guess, we have to expect `calledWithExactly` some arguments as we have above for second call and first call, because now I don't understand why it should call thrice

oleh-momot (approved):

ilc/client/isActiveFactory.spec.js
StyleT: Switched to `calledWithExactly `

ilc/client/utils.js
```diff
@@ -7,3 +7,16 @@ export function getSlotElement(slotName) {
 
     return appContainer || slot;
 }
+
+export function iterablePromise(iterable) {
```
nightnei, line 11: can't understand the method

nightnei (approved):

PR #216, opened 2020-10-15
Add settings to config, process trailing slash on server and client s…

oleh-momot:
…ides


registry/server/settings/interfaces/index.ts
```diff
@@ -2,33 +2,66 @@ import Joi from 'joi';
...
+                TrailingSlashValues.RedirectToBaseUrl,
+                TrailingSlashValues.RedirectToBaseUrlWithTrailingSlash,
+            ).required(),
+        },
+        {
+            is: Joi.valid(SettingKeys.AmdDefineCompatibilityMode, SettingKeys.AuthOpenIdEnabled),
+            then: Joi.boolean().strict().sensitive().required(),
+        }
+    ],
+    otherwise: Joi.string().required(),
```
StyleT, line 56: Let's use:
```
Joi.string().uri({
  scheme: [ /https?/ ],
  allowRelative: false
});
```

For `AuthOpenIdDiscoveryUrl ` and `BaseUrl `

registry/client/src/settings/Edit.js
```diff
@@ -0,0 +1,66 @@
...
+
+import * as validators from '../validators';
+import {choices, keys} from './dataTransform';
+
+const Title = ({record}) => {
+    return (<span>{record ? `Setting "${record.key}"` : ''}</span>);
+};
+
+const InputForm = (props) => {
+    switch(props.record.key) {
```
StyleT, line 18: I think we shouldn't hardcode any keys here. Addition of the new setting should be a matter of adding new row to DB and maybe adding new key to Joi schema on server side. So 1, max 2 update points.

StyleT: Think about the logic when server returns also: key type & possible values for select.

StyleT (changes requested):

registry/server/migrations/20201013140633_settings_default.ts
```diff
@@ -0,0 +1,63 @@
...
+        value: JSON.stringify('http://localhost:4001/'),
+        secured: true,
+    }, {
+        key: SettingKeys.TrailingSlash,
+        value: JSON.stringify(TrailingSlashValues.DoNothing),
+        secured: false,
+    }, {
+        key: SettingKeys.AuthOpenIdEnabled,
+        value: JSON.stringify(false),
+        secured: true,
```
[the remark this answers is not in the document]
StyleT: Note: 
scope=ilc|registry. 
secret: 0|1, only `AuthOpenIdClientSecret`

scope:
ilc: trailingSlash, amdDefineCompatibilityMode
registry: all other keys

ilc/common/trailingSlash.js
```diff
@@ -0,0 +1,32 @@
+const routerHasTo = {
+    doNothing: 'doNothing',
+    redirectToBaseUrl: 'redirectToBaseUrl',
+    redirectToBaseUrlWithTrailingSlash: 'redirectToBaseUrlWithTrailingSlash',
+};
+
+const processUrl = (registryConfig, url) => {
+    const trailingSlash = registryConfig.settings.trailingSlash;
```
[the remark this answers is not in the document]
oleh-momot: Yeah, It makes sense, I changed it

StyleT (commented):
I think you're missed to handle case when App uses pushHistory to change url

ilc/common/trailingSlash.js
```diff
@@ -0,0 +1,31 @@
...
+    doNothing: 'doNothing',
+    redirectToBaseUrl: 'redirectToBaseUrl',
+    redirectToBaseUrlWithTrailingSlash: 'redirectToBaseUrlWithTrailingSlash',
+};
+
+const processUrl = (trailingSlash, url) => {
+    const doesReqUrlEndWithTrailingSlash = url[url.length - 1] === '/';
+
+    switch (trailingSlash) {
+        case routerHasTo.doNothing: break;
```
StyleT, line 11: Move down to default

ilc/common/trailingSlash.js
```diff
@@ -0,0 +1,31 @@
+const routerHasTo = {
+    doNothing: 'doNothing',
+    redirectToBaseUrl: 'redirectToBaseUrl',
+    redirectToBaseUrlWithTrailingSlash: 'redirectToBaseUrlWithTrailingSlash',
+};
+
+const processUrl = (trailingSlash, url) => {
+    const doesReqUrlEndWithTrailingSlash = url[url.length - 1] === '/';
```
StyleT, line 8: I believe you missed to handle case (and test for it) when URL has query params and/or hash.

registry/client/src/settings/List.js
```diff
@@ -0,0 +1,88 @@
...
+            pagination={<Pagination />}
+        >
+            {isSmall
+                ? (
+                    <SimpleList
+                        primaryText={record => record.key}
+                        secondaryText={record => record.value}
+                    />
+                ) : (
+                    <Datagrid rowClick="edit" optimized>
```
StyleT, line 73: Could you attach a screenshot of this page?

registry/server/common/services/validateRequest.ts
```diff
@@ -29,6 +29,7 @@ const validateRequestFactory = (validationConfig: ValidationConfig[]) => async (
         ));
         next();
     } catch (e) {
+        console.error(e);
```
StyleT, line 32: Why do we need it here? Don't think it will be a useful info in production env...

registry/server/migrations/20201012152003_settings_meta.ts
```diff
@@ -0,0 +1,25 @@
+import * as Knex from "knex";
+import {
+    Scope,
+} from "../settings/interfaces";
+
+
+export async function up(knex: Knex): Promise<void> {
+    return knex.schema.table('settings', table => {
+        table.string('default', 50).notNullable().defaultTo(JSON.stringify(''));
```
StyleT, line 9: type for col `value` & `default` should match. So it must be TEXT here.

registry/server/routes/config.ts
```diff
@@ -72,6 +77,11 @@ router.get('/', async (req, res) => {
         return acc;
     }, {});
 
+    data.settings = preProcessResponse(settings).reduce((acc: {[key: string]: any}, setting: Setting) => ({
```
StyleT, line 80: Can we here treat dots in setting key as object levels.

So if we have `a.b` & `a.c` config endpoint should return:
```
{
  a: {b: 'XXX', c: 'XXX'}
}
```

registry/server/settings/services/SettingsService.ts
```diff
@@ -51,20 +40,21 @@ export class SettingsService {
...
         }
 
-        return value;
+        const defaultValue = JSON.parse(setting.default);
+
+        if (defaultValue !== '') {
+            return defaultValue;
+        }
+
+        return JSON.parse(setting.value);
```
StyleT, line 56: If this code is going to be executed - it will always return `''`

registry/client/src/settings/Edit.js
```diff
@@ -0,0 +1,58 @@
...
+        }
+        default: {
+            return (<TextInput source="value" />);
+        }
+    }
+};
+
+const MyEdit = (props) => {
+    return (
+        <Edit title={<Title />} undoable={false} {...props}>
```
StyleT, line 50: Could you attach a screenshot of this page?

StyleT (changes requested):

registry/server/migrations/20201012152003_settings_meta.ts
oleh-momot: Oh, I missed it, I looked at the `key`
I am going to change it

registry/client/src/settings/List.js
oleh-momot: Screenshot 2020-10-20 at 12 48 11

registry/client/src/settings/Edit.js
oleh-momot: Screenshot 2020-10-20 at 12 50 10

registry/server/routes/config.ts
oleh-momot: I decided to use `_.set()` here to have the result as you described

ilc/client/ClientRouter.js
```diff
@@ -134,6 +137,12 @@ export default class ClientRouter {
                 },
             });
         }
+
+        const processedUrl = this.#urlProcessor.process(this.#currentUrl);
```
oleh-momot, line 141: TODO: cover with tests

ilc/client/ClientRouter.js
```diff
@@ -134,6 +137,12 @@ export default class ClientRouter {
...
+        if (this.#currentUrl !== processedUrl) {
+            console.info(`ILC: Redirecting to "${processedUrl}"...`);
+            this.#location.replace(processedUrl);
```
oleh-momot, line 144: Need to discuss it
Because I am not sure about it

oleh-momot: Discussed
I removed it from here
Added to patched update state require of a function that would return a created instance of URL processor

ilc/client/getIlcUrlProcessor.js
```diff
@@ -0,0 +1,13 @@
+const UrlProcessor = require('../common/UrlProcessor');
+
+let urlProcessor;
+
+function getIlcUrlProcessor(trailingSlash) {
```
StyleT, line 5: Let's make this file more obvious. It should fetch necessary configs from `initIlcConfig` on it's own.

oleh-momot: Added

ilc/client/ClientRouter.js
```diff
@@ -150,7 +159,8 @@ export default class ClientRouter {
         const {specialRole} = this.#router.match(pathname);
 
         if (specialRole === null) {
-            this.#singleSpa.navigateToUrl(href);
+            const processedUrl = this.#urlProcessor.process(href);
```
StyleT, line 162: You don't need it here, see https://github.com/single-spa/single-spa/blob/d89173b6cadd4ba2e581a648a5ee0821488a71d8/src/navigation/navigation-events.js#L18

StyleT (changes requested):

ilc/client/ClientRouter.js
oleh-momot: It does not work without it unfortunately

oleh-momot: It was a mistake in the code
It works
I removed it from here

StyleT (approved):

registry/client/src/settings/List.js
nightnei: can we create one field (object) for `auth.openId`

nightnei:
can we create one field (object) for `auth.openId`?
https://github.com/namecheap/ilc/pull/216#discussion_r508363167

oleh-momot:
> can we create one field (object) for `auth.openId`?
> [#216 (comment)](https://github.com/namecheap/ilc/pull/216#discussion_r508363167)

Discussed
It is an improvement that we are going to add to Backlog

nightnei (approved):

PR #223, opened 2020-10-29
Feature/openid session impersonation

StyleT:


registry/server/auth.ts
```diff
@@ -255,5 +260,9 @@ async function getEntityWithCreds(provider: string, identifier: string, secret:
 
     delete user.secret;
```
oleh-momot, line 261: So, it does not need anymore if I got it correctly

registry/tests/auth.spec.ts
```diff
@@ -79,7 +81,7 @@ describe('Authentication / Authorization', () => {
                 })
                 .expect(200)
                 .expect('set-cookie', /connect\.sid=.+; Path=\/; HttpOnly/)
-                .expect('set-cookie', /ilc:userInfo=%7B%22identifier%22%3A%22root%22%2C%22role%22%3A%22admin%22%7D; Path=\//);
+                .expect('set-cookie', /ilc:userInfo=%7B%22authEntityId%22%3A1%2C%22identifier%22%3A%22root%22%2C%22role%22%3A%22admin%22%7D; Path=\//);
```
oleh-momot, line 84: Can we use URL encode and decode here and in other such places?
Because I guess, if something will be wrong it will take us an extra effort to understand exactly what it is wrong

oleh-momot (changes requested):

nightnei (approved):

registry/tests/auth.spec.ts
StyleT: fixed

registry/server/auth.ts
StyleT: fixed

oleh-momot (approved):

PR #225, opened 2020-11-10
Feature/versioning

StyleT:


registry/client/src/authEntities/dataTransform.js
```diff
@@ -0,0 +1,11 @@
+export function transformGet(entity) {
+
+}
+
+export function transformSet(entity, operation) {
+    if (operation === 'update') {
```
oleh-momot, line 6: In my opinion, it would be nice to have one object with operations and use it everywhere
because as I see operations are copy/pasted in each of the files

registry/client/src/versioning/List.js
```diff
@@ -0,0 +1,134 @@
...
+            filters={<MyFilter/>}
+            exporter={false}
+            perPage={25}
+            bulkActionButtons={false}
+        >
+            <Datagrid expand={<MyPanel />} >
+                <TextField sortable={false} source="id" />
+                <TextField sortable={false} source="entity_type" />
+                <TextField sortable={false} source="entity_id" />
+                <FunctionField label="Operation" render={record => record.data && record.data_after ? 'UPDATE' : record.data ? 'DELETE' : 'CREATE'} />
```
oleh-momot, line 123: So if you create an object that I mentioned [here](https://github.com/namecheap/ilc/pull/225/files#r521177193) you could use it here with `toUpperCase()`

registry/server/versioning/interfaces.ts
```diff
@@ -0,0 +1,31 @@
...
+    data_after: string|null;
+    created_by: string;
+    created_at: number;
+}
+export interface VersionRow extends VersionRowData {
+    id: string;
+}
+
+export interface VersionRowParsed extends VersionRow {
+    data: any|null;
```
oleh-momot, line 29: Are you positive that `null` is needed here?

registry/client/src/versioning/List.js
nightnei: I think that using here the same variable is a little bit incorrect since it's `View` with simple textual information which can be changed. So it's better to not mix presentation and logic inside a single variable.

registry/server/versioning/interfaces.ts
StyleT: not really, just a silly attempt to express that here we may have random object or null.. Do you know a better typing for this?

registry/client/src/authEntities/dataTransform.js
StyleT: fixed

registry/client/src/versioning/List.js
oleh-momot: Yeah, I agree with that

nightnei (approved):

registry/server/versioning/interfaces.ts
StyleT: fixed

oleh-momot (approved):

PR #230, opened 2020-11-20
Feature/i18n/refactoring and tests

StyleT:


ilc/client.js
```diff
@@ -5,38 +5,56 @@ import Router from './client/ClientRouter';
...
-const router = new Router(registryConf, state, singleSpa);
+
+const appErrorHandlerFactory = (appName, slotName) => {
+    return fragmentErrorHandlerFactory(registryConf, router.getCurrentRoute, appName, slotName);
+};
+
+const i18n = registryConf.settings.i18n.enabled
+    ? new I18n(registryConf.settings.i18n, singleSpa, appErrorHandlerFactory)
+    : null;
+const router = new Router(registryConf, state, i18n ? i18n.unlocalizeUrl : undefined, singleSpa);
```
nightnei, line 34: it's not critical, but will be better to put `undefined` arguments as last argument or as optional inside object. and in this case we won't set `undefiend` here.
https://github.com/namecheap/ilc/pull/230/files#diff-318e72ee3c278cb96c1bb2130002e8a9bcdebd2042222be2a5284e589a1bb75eR204

ilc/client/dispatchSynchronizedEvent.js
```diff
@@ -0,0 +1,62 @@
...
+    handlers.forEach(v => {
+        v.errorHandler = (...args) => errorHandler(v.actorId, ...args);
+        v.prepare = wrapWithPromise(v.prepare);
+        v.execute = wrapWithPromise(v.execute);
+    });
+
+    const preparationPromises = handlers.map(v => v.prepare(eventDetail));
+
+    return Promise.allSettled(preparationPromises).then(results => {
+        for (let ii = 0; ii < results.length; ii++) {
```
nightnei, line 29: why `ii`? 😁

ilc/client/dispatchSynchronizedEvent.js
```diff
@@ -0,0 +1,62 @@
...
+                ii--;
+            }
+        }
+
+        const executionPromises = handlers.map((v, i) => v.execute(eventDetail, results[i].value));
+
+        return Promise.allSettled(executionPromises);
+    }).then(results => {
+        results
+            .map((v, i) => {
```
nightnei, line 45: it's better to use `reduce` here

ilc/tests/helpers.js
```diff
@@ -0,0 +1,88 @@
+const _ = require('lodash');
```
nightnei, line 1: I propose to not use lodash if we can do it.
for example, here we can use `deepmerge` package.

nightnei (approved):

oleh-momot (approved):

PR #237, opened 2021-01-04
feat: add meta field to route

oleh-momot:


registry/server/migrations/20201228201314_route_meta.ts
```diff
@@ -0,0 +1,17 @@
+import * as Knex from "knex";
+
+
+export async function up(knex: Knex): Promise<any> {
+    return knex.schema.createTable('route_meta', table => {
```
StyleT, line 5: Instead of creating separate table - I would just add TEXT column to routes table.

StyleT (changes requested):

registry/server/migrations/20201228201314_route_meta.ts
oleh-momot: Updated

StyleT (approved):

PR #246, opened 2021-01-22
feat: GuardManager & common PluginManager

oleh-momot:


ilc/server/pluginManager/PluginManager.js
```diff
@@ -53,5 +58,15 @@ module.exports = class PluginManager {
             console.info(`Enabling plugin "${manifest.name}" of type "${pSource.type}"...`);
             this.#plugins[pSource.type] = pSource;
         }
+
+        this.#plugins[TYPES.transitionHooks] = {
```
oleh-momot, line 62: Mocked `transitionHooks` plugin

ilc/client/GuardManager.js
```diff
@@ -0,0 +1,112 @@
...
+    }
+};
+
+let guardManager;
+
+module.exports = () => {
+    if (guardManager === undefined) {
+        const router = getIlcRouter();
+        const registryConf = getIlcConfig();
+        const transitionHooksPlugin = {
```
oleh-momot, line 63: Mocked `transitionHooks` plugin

ilc/build/webpack.js
```diff
@@ -35,6 +35,11 @@ module.exports = {
             header: () => fs.readFileSync(path.resolve(__dirname, '../public/system.js')),
             afterOptimizations: true,
         }),
+        new WrapperPlugin({
+            test: /\.js$/,
+            header: () => fs.readFileSync(path.resolve(__dirname, '../public/patchEvents.js')),
```
oleh-momot, line 40: It is needed to monkey-patch `popstate`, `hashchange` events before single-spa will patch them

ilc/client/ClientRouter.js
```diff
@@ -1,8 +1,9 @@
 import deepmerge from 'deepmerge';
 
-import * as Router from '../common/router/Router';
 import * as errors from '../common/router/errors';
 
+const getIlcRouter = require('./getIlcRouter');
```
oleh-momot, line 5: It is needed to use Router as a singleton

ilc/client/GuardManager.js
```diff
@@ -0,0 +1,112 @@
...
+        const hooks = this.#transitionHooksPlugin.getTransitionHooks();
+
+        if (route.specialRole == null) {
+            for (const hook of hooks) {
+                const action = hook({
+                    route: {
+                        ...route,
+                        url,
+                    },
+                    navigate: this.#navigateToUrl,
```
oleh-momot, line 42: I am not positive that we have to provide just navigateToUrl method from single-spa
Probably, we have to check that an URL that would be provided from a hook single-spa has

ilc/client/navigationEvents/setupEvents.js
```diff
@@ -18,7 +26,24 @@ function patchedUpdateState(updateState) {
...
+function handlePopState(event) {
+    const oldUrl = currentUrl;
+    const newUrl = (currentUrl = window.location.href);
+
+    if (event.singleSpa || shouldNavigateTo(newUrl)) {
+        window.callCapturedEventListeners(arguments);
+        return;
+    }
+
+    window.history.replaceState(null, undefined, oldUrl);
```
oleh-momot, line 40: Now, I have decided to use `history.replaceState` here to change the current `location.href` because we do not have an opportunity to check what exactly button would be clicked by a user: forward, back, or an entry in History.

ilc/server/i18n.js
```diff
@@ -59,21 +59,6 @@ const onRequestFactory = (i18nConfig, i18nParamsDetectionPlugin) => async (req,
     }
 }
 
-/**
```
oleh-momot, line 62: I have decided to move it to common because it is used on both server and client sides

ilc/package.json
```diff
@@ -11,10 +11,11 @@
     "test:client": "npm run build:systemjs && cross-env NODE_ENV=test karma start",
     "test:client:watch": "npm run test:client -- --single-run=false --auto-watch=true",
     "start": "node --max-http-header-size 30000 server/index.js",
-    "dev": "npm run build:polyfills && cross-env NODE_ENV=development nodemon --ignore './public/' --max-http-header-size 30000 server/index.js",
-    "build": "rimraf public && npm run build:systemjs && npm run build:client && npm run build:polyfills",
+    "dev": "npm run build:polyfills && npm run build:patch-events && cross-env NODE_ENV=development nodemon --ignore './public/' --max-http-header-size 30000 server/index.js",
+    "build": "rimraf public && npm run build:systemjs && npm run build:patch-events && npm run build:client && npm run build:polyfills",
     "build:client": "webpack --config build/webpack.js --progress --hide-modules",
     "build:systemjs": "node ./systemjs/build.js",
+    "build:patch-events": "babel ./client/navigationEvents/patchEvents.js --out-file ./public/patchEvents.js && uglifyjs ./public/patchEvents.js --output ./public/patchEvents.js --compress --mangle --enclose",
```
oleh-momot, line 18: It needs to compile `patchEvents.js` to ES5 and minify it

ilc/client/navigationEvents/patchEvents.js
```diff
@@ -0,0 +1,61 @@
...
+function patchedRemoveEventListener(eventName, fn) {
+    if (typeof fn === "function" && routingEventsListeningTo.indexOf(eventName) >= 0) {
+        capturedEventListeners[eventName] = capturedEventListeners[eventName].filter((listener) => listener !== fn);
+        return;
+    }
+
+    return window.unsetEventListener.apply(this, arguments);
+}
+
+window.setEventListener = window.addEventListener;
```
oleh-momot, line 55: In this way, we can save the origin method and we can use it to set our event handlers to `postate` and `hashchange` events that can call captured event handlers if it is needed

ilc/client/GuardManager.js
StyleT: Pls move it to separate file & load via some PluginManager, as we do it at server side

ilc/client/GuardManager.js
```diff
@@ -0,0 +1,112 @@
...
+        this.#registryConf = registryConf;
+        this.#router = router;
+        this.#unlocalizeUrl = unlocalizeUrl;
+        this.#navigateToUrl = navigateToUrl;
+        this.#transitionHooksPlugin = transitionHooksPlugin;
+    }
+
+    shouldNavigateTo(url) {
+        const pathname = new URL(url, 'http://hack').pathname;
+        const route = this.#router.match(this.#unlocalizeUrl(this.#registryConf.settings.i18n, pathname));
```
StyleT, line 32: This looks like a bad design decision... We use common/Router here directly & totally ignore existence of the ClientRouter... So we expect `match()` fm here return same result as ClientRouter will get - but this may not be true as we already have some "preparation" code (unlocalizeUrl) which is copy pasted here from ClientRouter.

ilc/server/app.js
```diff
@@ -24,6 +27,15 @@ module.exports = (registryService, pluginManager) => {
         await i18nOnRequest(req, reply);
     });
 
+    if (transitionHooksPlugin !== null) {
+        app.addHook('preHandler', async (req, reply) => {
```
StyleT, line 31: Let's discuss... Changes from my PR may help to do this part of logic in a better way.

StyleT (changes requested):

ilc/build/webpack.js
oleh-momot: Because `setupEvents.js` uses dependencies from our bundle
`patchEvents.js` should not use any dependencies because it is not a part of client.js bundle itself and this file does not have access to them

ilc/client/GuardManager.js
```diff
@@ -0,0 +1,112 @@
+const navigateToUrl = require('single-spa').navigateToUrl;
```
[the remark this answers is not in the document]
oleh-momot: Because we require `GuardManager` in `popstate`, `hashchange`, change state event handlers as a singleton

ilc/client/transitionHooksPlugin.js
```diff
@@ -0,0 +1,45 @@
+import actionTypes from '../common/guard/actionTypes';
+
+export default {
```
oleh-momot, line 3: Mocked `transitionHooks` plugin

ilc/client/GuardManager.js
StyleT: still relevant

ilc/client.js
```diff
@@ -14,7 +14,10 @@ import AsyncBootUp from './client/AsyncBootUp';
 import IlcAppSdk from 'ilc-sdk/app';
 import I18n from './client/i18n';
 import {makeAppId} from './common/utils';
+import UrlProcessor from './common/UrlProcessor';
```
StyleT, line 17: For some reason I don't see this file in PR

ilc/client/GuardManager.js
```diff
@@ -0,0 +1,32 @@
+import actionTypes from '../common/guard/actionTypes';
+
+export default class GuardManager {
+    #router;
+    #transitionHooksPlugin;
+
+    constructor(router, transitionHooksPlugin) {
```
StyleT, line 7: I would pass plugin manager instead of transitionHooksPlugin here

ilc/client/navigationEvents/setupEvents.js
```diff
@@ -1,13 +1,84 @@
...
+    }
+
+    return removeEventListener.apply(this, arguments);
+}
+
 function fireRoutingEvent() {
     window.dispatchEvent(new CustomEvent('ilc:before-routing'));
 }
 
+function setupNavigationHook(hook) {
```
StyleT, line 65: Could we have multiple hooks here?

ilc/client.js
oleh-momot: I have created this file previously when I worked on leading trailing slash task

ilc/client/GuardManager.js
oleh-momot: Updated

ilc/client.js
```diff
@@ -29,17 +32,26 @@ const appErrorHandlerFactory = (appName, slotName) => {
     return fragmentErrorHandlerFactory(registryConf, router.getCurrentRoute, appName, slotName);
 };
 
+const pluginManager = new PluginManager(require.context('./node_modules', true, /ilc-plugin-.*\/dist\/browser\.js$/));
```
oleh-momot, line 35: I have decided to use [`require.context`](https://webpack.js.org/guides/dependency-management/#requirecontext) to find every plugin and include them into `client.js` bundle
In this way, we can use common shared logic for PluginManager because the only thing that we have to change on the server side is to use require.context as a npm package

ilc/client/GuardManager.js
oleh-momot: Moved to the folder `ilc-plugin-transition-hooks` that we can move to `node_modules` into any folder (e.g. `node_modules/@test/ilc-plugin-transition-hooks`) for testing

ilc/client/navigationEvents/setupEvents.js
oleh-momot: Updated, I have added the ability to do it

ilc/server/app.js
oleh-momot: Updated

ilc/client/GuardManager.js
```diff
@@ -0,0 +1,36 @@
...
+
+        const route = this.#router.match(url);
+
+        if (route.specialRole === null) {
+            for (const hook of this.#transitionHooksPlugin.getTransitionHooks()) {
+                const action = hook({
+                    route: {...route, url},
+                    navigate: this.#router.navigateToUrl,
+                });
+
```
oleh-momot, line 27: TODO
Also, need to handle `redirect` action type here

ilc/client/GuardManager.js
```diff
@@ -0,0 +1,36 @@
...
+export default class GuardManager {
+    #router;
+    #transitionHooksPlugin;
+
+    constructor(router, pluginManager) {
+        this.#router = router;
+        this.#transitionHooksPlugin = pluginManager.getTransitionHooksPlugin();
+    }
+
+    hasAccessTo(url) {
```
oleh-momot, line 12: TODO
Show 500 error page when hook throws an error

ilc/server/GuardManager.js
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        const route = req.raw.router.getRoute();
+        const hooks = this.#transitionHooksPlugin.getTransitionHooks();
+
+        if (hooks.length === 0) {
+            return hasAccess;
+        }
+
+        const actions = await Promise.all(hooks.map((hook) => hook({
```
StyleT, line 24: Some try/catch with explicit error is necessary here...

ilc/server/GuardManager.js
```diff
@@ -0,0 +1,40 @@
+const actionTypes = require('../common/guard/actionTypes');
+
+class GuardManager {
+    #transitionHooksPlugin;
+
+    constructor(pluginManager) {
+        this.#transitionHooksPlugin = pluginManager.getTransitionHooksPlugin();
+    }
+
+    async hasAccessTo(req, res) {
```
StyleT, line 10:
```suggestion
    async redirectTo(req):string|null {
```

StyleT: And in `app.js`:

```
const redirectTo = await guardManager.redirectTo(req.raw);
if (redirectTo) {
            res.redirect(redirectTo);
            return;
        }
```

ilc/server/i18n.spec.js
```diff
@@ -53,12 +53,6 @@ describe('i18n', () => {
     afterEach(() => {
         i18nParamsDetectionPlugin.detectI18nConfig.reset();
         pluginManager.getI18nParamsDetectionPlugin.reset();
-        pluginManager.getReportingPlugin.reset();
-    });
-
-    after(() => {
```
StyleT, line 59: Why have you removed this?

StyleT:
I would expect Route Guards to be called one by one and the nect one should be called only if all previous ones have returned "noop" (ex: null).

StyleT (changes requested):

ilc/client.js
```diff
@@ -19,16 +23,30 @@ const appErrorHandlerFactory = (appName, slotName) => {
...
 const i18n = registryConf.settings.i18n.enabled
     ? new I18n(registryConf.settings.i18n, {...singleSpa, triggerAppChange}, appErrorHandlerFactory)
     : null;
 const router = new Router(registryConf, state, i18n ? i18n.unlocalizeUrl : undefined, singleSpa);
+const guardManager = new GuardManager(router, pluginManager, internalErrorHandler);
+const urlProcessor = new UrlProcessor(registryConf.settings.trailingSlash);
+
+addNavigationHook((state) => {
+    if (state.nextUrl) {
```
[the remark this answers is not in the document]
oleh-momot: I have added it because it is in history update state methods can be undefined in some cases https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState (URL is optional)

ilc/client/navigationEvents/setupEvents.js
```diff
@@ -1,24 +1,145 @@
...
+}
+
+function callNavigationHooks(url) {
+    let state = {
+        navigationShouldBeCanceled: false,
+        nextUrl: url,
+    };
+
+    for (const hook of hooks) {
+        const nextState = hook(state);
```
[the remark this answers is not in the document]
StyleT: `setupEvents` exposes `setErrorHandler` which will be registered in `setupErrorHandlers`.

By default - `setErrorHandler` only logs to console.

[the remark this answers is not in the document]
oleh-momot: Updated

ilc/server/GuardManager.js
oleh-momot: Updated

ilc/client/GuardManager.js
```diff
@@ -0,0 +1,66 @@
...
+        const hooks = this.#transitionHooksPlugin.getTransitionHooks();
+
+        if (hooks.length === 0) {
+            return true;
+        }
+
+        for (const hook of hooks) {
+            try {
+                const action = hook({
+                    route: {...route, url},
```
StyleT, line 35: Here in `route` object you need to pass only `url` & `meta` keys.

StyleT: Same for SSR interface

StyleT (changes requested):

ilc/client/GuardManager.js
```diff
@@ -0,0 +1,66 @@
...
+                    navigate: this.#router.navigateToUrl,
+                });
+
+                if (action.type === actionTypes.stopNavigation) {
+                    return false;
```
StyleT, line 40: Add here log to console: `ILC: Stopped navigation due to the Route Guard with index #XXX`

ilc/client/GuardManager.js
```diff
@@ -0,0 +1,66 @@
...
+                }
+
+                if (action.type === actionTypes.redirect) {
+                    // Need to add redirect callback to queued tasks
+                    // because it should be executed after micro tasks that can be added after the end of this method
+                    setTimeout(() => this.#router.navigateToUrl(action.newLocation));
```
StyleT, line 46: Add here log to console: ILC: Redirect from "%location.href%" to "%location.href%" due to the Route Guard with index #XXX

StyleT: Same at SSR, but ensure that at server side you write to logger instead of "console"

ilc/client/GuardManager.js
oleh-momot: Updated

ilc/client/GuardManager.js
oleh-momot: Updated

ilc/client/GuardManager.js
oleh-momot: Updated

ilc/client/navigationEvents/setupEvents.js
```diff
@@ -134,6 +134,17 @@ function callNavigationHooks(url) {
...
 function patchedUpdateState(updateState) {
     return function (state, title, url, ...rest) {
+        /**
+         * Need to check "url" argument because dependencies as "@mapbox/scroll-restorer" provides "window.location" when using History API
+         * This parameter should be a string or undefined due to the documentation
+         *
+         * @see {@link https://developer.mozilla.org/en-US/docs/Web/API/History_API}
+         * @see {@link https://github.com/mapbox/scroll-restorer/blob/main/index.js#L65}
+         */
+        if (typeof url !== 'string' && window.location === url) {
```
StyleT, line 144: Change to 
```
if (typeof url === 'object' && url.pathname) {
    url = url.pathname + url.search + url.hash;
}

I believe with current approach it will cut search string & hash off the URL during scroll.
```

oleh-momot: Updated

StyleT (approved):

nightnei (approved):

PR #263, opened 2021-02-23
fix: add a seed to settings table

oleh-momot:


registry/server/seeds/05_settings.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export async function seed(knex: Knex): Promise<any> {
+    const settings: Pick<Setting, 'key' | 'value'>[] = [];
+
+    if (overrideConfigTrustedOrigins) {
+        settings.push({
+            key: SettingKeys.OverrideConfigTrustedOrigins,
+            value: overrideConfigTrustedOrigins,
+        });
+    }
```
StyleT, line 18: Add else here which resets value to it's default state

StyleT (changes requested):

StyleT (approved):

PR #271, opened 2021-03-10
Feature/parcels

StyleT:


ilc/client/BundleLoader.js
```diff
@@ -0,0 +1,82 @@
+import crashIlc from './errorHandler/crashIlc';
+
+export class BundleLoader {
+
+    #cache = new WeakMap();
```
oleh-momot, line 5: Please add WeakMap into the URL that we use to build polyfills with the help of `build:polyfills` script

oleh-momot (changes requested):

nightnei (approved):

ilc/client/BundleLoader.js
StyleT: I they're supported even in IE 11. Are you sure we need polyfill for them?

oleh-momot: Discussed

oleh-momot (approved):

PR #275, opened 2021-03-23
fix: exclude / from regexp to search browser plugins

oleh-momot:


nightnei (approved):

ilc/client.js
```diff
@@ -32,7 +32,7 @@ const appErrorHandlerFactory = (appName, slotName) => {
     return fragmentErrorHandlerFactory(registryConf, router.getCurrentRoute, appName, slotName);
 };
 
-const pluginManager = new PluginManager(require.context('./node_modules', true, /ilc-plugin-.*\/browser\.js$/));
+const pluginManager = new PluginManager(require.context('./node_modules', true, /ilc-plugin-[^/]*\/browser\.js$/));
```
nightnei, line 35: not critical and it can't cause any bug, but a little bit better to use `+` instead of `*` here

oleh-momot: Updated

StyleT (approved):

PR #285, opened 2021-04-20
Feature/support a few domains

nightnei:


StyleT (changes requested):
- Routes list: 
    - domainId -> Domain
    - There should be a way to see only routes without domain
- http://localhost:4001/#/route/create 
    - Put Domain selection above metadata. 
    - Also add field explanation
- Main menu: put router domains right after Routes 
- Domain creation/update: 
    - 500 template should be required
    - Domain name should be validated
- Unclear error when trying to delete domain with related routes. Err message should say why it's impossible

nightnei:
> * Routes list:
>   
>   * domainId -> Domain
>   * There should be a way to see only routes without domain
> * http://localhost:4001/#/route/create
>   
>   * Put Domain selection above metadata.
>   * Also add field explanation
> * Main menu: put router domains right after Routes
> * Domain creation/update:
>   
>   * 500 template should be required
>   * Domain name should be validated
> * Unclear error when trying to delete domain with related routes. Err message should say why it's impossible

done

StyleT (approved):

PR #293, opened 2021-05-06
chore: doc for multi-domains

nightnei:


docs/multi-domains.md
```diff
@@ -0,0 +1,37 @@
+# Multi-domains
+
+ILC supports multi-domains, which means that you don't need to roll out personal instances of ILC for each of them, you just can use one instance for all of them.
```
StyleT, line 3:
```suggestion
ILC can handle requests coming from multiple domains, so that you don't need to roll out individual instances of the ILC for every domain, you just can use one instance for all of them.  
```

docs/multi-domains.md
```diff
@@ -0,0 +1,37 @@
+# Multi-domains
```
StyleT, line 1: Here we also need to mention the following:
- ILC detects domain from the request by using `Host` HTTP header
- Every registered domain has own set of routes. They do not overlap.
- If request is going to the domain which is absent in **Router domains** - routes w/o assigned domain will be used for routing.

StyleT (changes requested):

docs/multi-domains.md
nightnei: done

StyleT (approved):

PR #295, opened 2021-05-12
feat: ability to filter ILC Routes in registry by route string

StyleT:


registry/client/src/appRoutes/List.js
```diff
@@ -72,6 +73,11 @@ const ListFilter = ({ routerDomain, ...props }) => {
                     />
                 : null
             }
+            {
+                !props.filterValues.showSpecial
+                ? <TextInput label="Route prefix" source="routePrefix" alwaysOn resettable/>
```
nightnei, line 78: why do you use word "prefix"? I think label can be just "Route" and source="route"

nightnei, line 78:
```suggestion
                ? <TextInput label="Route" source="route" alwaysOn resettable/>
```

registry/server/appRoutes/routes/getAppRoutes.ts
```diff
@@ -16,6 +16,10 @@ const getAppRoutes = async (req: Request, res: Response) => {
         query.where('routes.route', 'like', `${SPECIAL_PREFIX}%`);
     } else {
         query.whereNot('routes.route', 'like', `${SPECIAL_PREFIX}%`);
+
+        if (filters.routePrefix !== undefined) {
+            query.where('routes.route', 'like', `${filters.routePrefix}%`)
```
nightnei, line 21: and here it's better to allow whole search with "%" at the beginning

```suggestion
            query.where('routes.route', 'like', `%${filters.route}%`)
```

nightnei (changes requested):

registry/client/src/appRoutes/List.js
nightnei: why do you use word "prefix"? I think label can be just "Route" and source="route"

nightnei (approved):

registry/client/src/appRoutes/List.js
StyleT: To emphasise that it is a prefix based search. Ex:
/foo maches: /foo, /foo/ & /foo/bar but doesn't match /test/foo

PR #298, opened 2021-05-20
chore: added tests for fragment hooks

ifree92:


StyleT (approved):

ilc/server/tailor/fragment-hooks.spec.js
```diff
@@ -0,0 +1,175 @@
...
+const {PassThrough} = require('stream');
+
+describe('fragment-hooks', () => {
+    describe('insertEnd', () => {
+        it('should not write to stream', () => {
+            const mockStream = new PassThrough();
+            const fragmentAttrs = getFragmentAttributes();
+            const headers = {};
+
+            insertEnd(mockStream, fragmentAttrs, headers, 0);
```
nightnei, line 13: looks like "index" isn't used inside "insertStart", probably we can remove it from module and from tests too

offtop: I think it's better to create a variable before providing raw parameter to a function, since it's unreal to understand what is the fourth parameter, e.g.
```
const indexOfSomething = 0;
insertEnd(mockStream, fragmentAttrs, headers, indexOfSomething);
```

nightnei (changes requested):

nightnei (approved):

PR #302, opened 2021-05-25
chore: added tests for client WrapApp class

ifree92:


ilc/client/WrapApp.spec.js
```diff
@@ -0,0 +1,291 @@
+import WrapApp from './WrapApp';
+import chai from 'chai';
+import sinon from 'sinon';
+
+describe('WrapApp', () => {
```
StyleT, line 5: 1. You're missing test cases for unmount (when we have wrapper or app rendered).
2. I would try to write utility function to check state of all callbacks at once. Ex. `expectBallbacksToBe(callbacks, 'mounted')`
3. Separately we need to test the fact that App Wrapper always receives `getCurrentBasePath() = '/'`, unique appId and it's own props while wrapped app receives "own" props from registry.
4. We also need to check that class also works with callbacks provided as array of functions. See `#canonizeCallbacks`

StyleT (changes requested):

ifree92:
@StyleT I have made changes regarding your comments.
Much detalised each case and allocated tests with "arrayed" callbacks as a different tests.

Probably, would be good to group them either:
- by general scenarios: boostrap/mount/unmount wrapper, bootstrap/mount/unmount wrapped app, dynamic render wrapped app after wrapper has been rendered
- or by callbacks approaches, e.g.: normal and "arrayed".

For now leaved as is.

Other moment is about expectations check of arrayed callbacks arguments. I have hardcoded them manually to be able to track the error on exact line. If you prefer to have there an additional abstraction layer - I will do, let me know.

StyleT (approved):

nightnei (approved):

PR #303, opened 2021-05-25
chore: added coverage text reporting npm command for ilc

ifree92:
Just useful thing upon development to check the general coverage via terminal

### DIFF
--- a/ilc/package.json
+++ b/ilc/package.json
@@ -8,6 +8,7 @@
     "test:ci": "npm run test -- --forbid-only",
     "test:watch": "npm run test -- --watch",
     "test:coverage": "cross-env NODE_ENV=test nyc mocha",
+    "test:coverage-text": "cross-env NODE_ENV=test nyc --reporter=text mocha",
     "test:client": "npm run build:systemjs && cross-env NODE_ENV=test karma start",
     "test:client:watch": "npm run test:client -- --single-run=false --auto-watch=true",
     "start": "node --max-http-header-size 30000 server/index.js",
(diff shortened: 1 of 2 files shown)


ilc/package.json
```diff
@@ -8,6 +8,7 @@
     "test:ci": "npm run test -- --forbid-only",
     "test:watch": "npm run test -- --watch",
     "test:coverage": "cross-env NODE_ENV=test nyc mocha",
+    "test:coverage-text": "cross-env NODE_ENV=test nyc --reporter=text mocha",
```
StyleT, line 11: Why would you need it?

ifree92: @StyleT Probably that might be useful for someone to check the coverage quickly and be sure about whether they have covered all mandatory lines.

I'm using that for analysis to be sure whether I have missed covering some conditions/lines or maybe that doesn't make sense in some context.

image

StyleT: Could you instead change config of the `npm run test:coverage` so it would produce text report alongside with html? Should fulfil your needs, enhance CI output while ensuring that we don't have "unused" scripts.

nightnei: as for me, it looks like a useful reminder to cover all files with tests :)

ifree92: @StyleT agreed.
Just was concerning to have some redundant output upon CI building.

I've made a new commit with replaced text reporter output into default coverage reporting configuration.
Could you please have a look

StyleT (approved):

nightnei (approved):

PR #305, opened 2021-05-31
feat: added ability to conveniently get hostname from client/server API of the Transition hooks plugin

StyleT:


ifree92 (approved):

ilc/client/GuardManager.js
```diff
@@ -34,10 +34,12 @@ export default class GuardManager {
                     route: {
                         meta: route.meta,
                         url: route.reqUrl,
+                        hostname: window.location.hostname,
```
nightnei, line 37: it's minor but I think it's a little bit better to use `host` instead of `hostname`. what do you think?

nightnei (approved):

ilc/client/GuardManager.js
StyleT: Nope :) http://bl.ocks.org/abernier/3070589 I would work inconsistently with SSR then.

PR #313, opened 2021-07-02
feat: Added support of the relative (to spaBundle URL) URLs in appAss…

StyleT:
…ets overrides


ilc/server/tailor/fragment-hooks.js
```diff
@@ -55,6 +57,16 @@ function insertEnd(stream, attributes, headers, index) {
     // disabling default TailorX behaviour
 }
 
+function fixUri(fragmentAttrs, uri) {
+    if (!isRelativeUrl(uri)) {
```
nightnei, line 61: looks like we can avoid calling isRelativeUrl, look at the example here https://developer.mozilla.org/ru/docs/Web/API/URL/URL#%D0%BF%D1%80%D0%B8%D0%BC%D0%B5%D1%80%D1%8B

```
var k = new URL('http://www.example.com', 'https://developers.mozilla.com');
// Creates a URL pointing to 'http://www.example.com/'
```

nightnei (approved):

nightnei (approved):

ilc/server/tailor/fragment-hooks.js
StyleT: done

PR #320, opened 2021-09-02
feat: shared libs

nightnei:


ilc/client/ilcConfig.js
```diff
@@ -12,13 +12,12 @@ export default function () {
 
     registryConf = JSON.parse(confScript.innerHTML);
 
-    document.head.appendChild(getSystemjsImportmap(registryConf.apps));
+    document.head.appendChild(getSystemjsImportmap(registryConf));
```
StyleT, line 15: Let's pass apps & sharedLibs explicitly as 2 params

registry/server/sharedLibs/routes/createSharedLib.ts
```diff
@@ -0,0 +1,34 @@
...
+const validateRequestBeforeCreateSharedLib = validateRequestFactory([{
+    schema: sharedLibSchema,
+    selector: 'body',
+}]);
+
+const createSharedLib = async (req: Request, res: Response): Promise<void> => {
+    const sharedLib = req.body;
+
+    // "dependencies" comes from assets-discovery.json, but we don't have column dependencies at the current moment
+    delete sharedLib.dependencies;
```
StyleT, line 23: Let's use white list approach here instead of black list

registry/server/sharedLibs/routes/updateSharedLib.ts
```diff
@@ -0,0 +1,55 @@
...
+        selector: 'body'
+    },
+]);
+
+const updateSharedLib = async (req: Request<UpdateSharedLibRequestParams>, res: Response): Promise<void> => {
+    const sharedLib = req.body;
+    const sharedLibName = req.params.name;
+
+    // "dependencies" comes from assets-discovery.json, but we don't have column dependencies at the current moment
+    delete sharedLib.dependencies;
```
StyleT, line 38: Let's use white list approach here instead of black list

registry/client/src/sharedLibs/Edit.js
```diff
@@ -0,0 +1,92 @@
...
+    return errors;
+};
+
+const InputForm = ({mode = 'edit', ...props}) => {
+    return (
+        <TabbedForm {...props}>
+            <FormTab label="Summary">
+                {mode === 'edit'
+                    ? <TextField source="name" />
+                    : <TextInput source="name" fullWidth />}
```
StyleT, line 51: I think we shouldn't allow change of name since it's a primary key and ID of the entity

registry/client/src/sharedLibs/Edit.js
```diff
@@ -0,0 +1,92 @@
...
+
+    return errors;
+};
+
+const InputForm = ({mode = 'edit', ...props}) => {
+    return (
+        <TabbedForm {...props}>
+            <FormTab label="Summary">
+                {mode === 'edit'
+                    ? <TextField source="name" />
```
StyleT, line 50: We should display name with `@sharedLibrary/` prefix

registry/client/src/sharedLibs/List.js
```diff
@@ -0,0 +1,48 @@
...
+            empty={<Empty />}
+        >
+            {isSmall ? (
+                <SimpleList
+                    primaryText={record => record.name}
+                    secondaryText={record => record.spaBundle}
+                />
+            ) : (
+                <Datagrid rowClick="show" optimized>
+                    <TextField source="name" />
```
StyleT, line 37: We should display name with `@sharedLibrary/` prefix

StyleT (changes requested):

registry/client/src/sharedLibs/Edit.js
nightnei: but here we use TextFIELD for "edit" and "TextINPUT" for "view",
or I missed something?

StyleT (approved):

PR #322, opened 2021-09-09
chore: add global method getAllSharedLibs

nightnei:


StyleT (approved):

ilc/client.js
```diff
@@ -64,6 +64,7 @@ const parcelApi = new ParcelApi(registryConf, bundleLoader, window.ILC.getAppSdk
 window.ILC.mountRootParcel = singleSpa.mountRootParcel;
 window.ILC.loadApp = bundleLoader.loadAppWithCss.bind(bundleLoader); // Internal API for Namecheap, not for public use
 window.ILC.importParcelFromApp = parcelApi.importParcelFromApp;
+window.ILC.getAllSharedLibNames = () => Object.keys(registryConf.sharedLibs);
```
StyleT, line 67: Could you make it async also? In future we may load this info lazily.

PR #328, opened 2021-11-16
Fix using styles and removing for unused apps

nightnei:


ilc/client/utils.js
```diff
@@ -8,20 +8,22 @@ export function getSlotElement(slotName) {
     return appContainer || slot;
 }
 
-export function prependSpaCallback(spaCallbacks, type, callback) {
+export function prependSpaCallbacks(spaCallbacks, newCallbacks) {
```
StyleT, line 11: Should be covered with tests

ilc/client/TransactionManager.js
```diff
@@ -106,9 +108,15 @@ export class TransactionManager {
         observer.observe(targetNode, { childList: true, subtree: true, attributeFilter: ['style'] });
     };
 
-    #renderFakeSlot = slotName => {
+    #renderFakeSlot = (slotName, appName) => {
         const targetNode = getSlotElement(slotName);
         const clonedNode = targetNode.cloneNode(true);
+
+        const cssBundleLink = document.querySelector(`link[data-fragment-id="${appName}"]`);
```
StyleT, line 115: I'm afraid that in case you have same app rendered twice on the page (ex: header/footer) - you would eventually remove links that you shouldn't be touching.

StyleT: Also I don't like that now `handlePageTransaction()` knows about appName. Looks like abstraction leakage a bit

ilc/client/WrapApp.js
```diff
@@ -89,13 +96,16 @@ export default class WrapApp {
...
+
+        this.#transactionManager.handlePageTransaction(slotName, slotWillBe.rerendered, this.#wrapperName);
 
         await wrapperCallbacks.unmount(props);
 
         if (this.#appState < APP_STATES.bootstrapped) {
             await appCallbacks.bootstrap(props);
         }
+
+        await this.#bundleLoader.loadCssByAppName(this.#appName);
```
StyleT, line 108: I don't think that it should be here... If we need to load CSS on app mount - it should be happening inside `appCallbacks.mount(props)`

ilc/client/TransactionManager.js
```diff
@@ -121,8 +129,16 @@ export class TransactionManager {
     #onAllSlotsLoaded = () => {
         this.#fakeSlots.forEach(node => node.remove());
         this.#fakeSlots.length = 0;
+
         this.#hiddenSlots.forEach(node => node.style.display = '');
         this.#hiddenSlots.length = 0;
+
+        this.#cssToRemove.forEach(link => {
+            window.System.delete(link.href);
```
StyleT, line 137: Why would you remove CSS from SystemJS?

ilc/client/BundleLoader.js
```diff
@@ -40,21 +40,36 @@ export class BundleLoader {
...
-        return this.#systemJs.import(url).catch(err => { //TODO: inserted <link> tags should have "data-fragment-id" attr. Same as Tailor now does
+    loadCss(url, appName) {
+        if (!url) {
+            return Promise.resolve();
+        }
+
+        return this.#systemJs.import(url)
+        .then(() => {
+            const currentLink = document.querySelector(`link[href="${url}"]`)
+            currentLink.setAttribute('data-fragment-id', appName)
```
StyleT, line 57: Minor: To me it looks like something that should have been done at SSR/inside System.js.

StyleT (changes requested):

PR #368, opened 2022-02-15
feat: add better handling of timeout errors - now error will contain …

b1ff:
…explicitly timeout and number of milliseconds from the request start.

### DIFF
--- a/ilc/server/tailor/request-fragment.spec.js
+++ b/ilc/server/tailor/request-fragment.spec.js
@@ -6,6 +6,7 @@ const requestFragmentSetup = require('./request-fragment');
 const ServerRouter = require('./server-router');
 const { getRegistryMock } = require('../../tests/helpers');
 const { getFragmentAttributes } = require('../../tests/helpers');
+const errors = require('./errors');
 
 describe('request-fragment', () => {
     /**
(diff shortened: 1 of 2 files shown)


ilc/server/tailor/request-fragment.spec.js
```diff
@@ -205,4 +206,51 @@ describe('request-fragment', () => {
...
+            })
+            .delay(timeoutMs + 20)
+            .reply(200);
+
+        try {
+            await requestFragment(attributes.url, attributes, request);
+            mockRequestScope.done();
+            chai.expect.fail('This code should not be reached, because error expected to be thrown above');
+        } catch (e) {
+            chai.expect(e).to.be.an.instanceof(errors.FragmentRequestError);
```
nightnei, line 252: looks like the test isn't valid, since if request isn't aborted due to timeout, then these two expects won't be used

nightnei (changes requested):

ilc/server/tailor/request-fragment.spec.js
nightnei: oh, I see, you made workaround with the help of `chai.expect.fail` in line 250

nightnei (approved):

PR #378, opened 2022-03-16
Increase db text types

blackrabbit99:
Registry stores templates in DB with type text which by default has 60kb capacity. 
The issue we faced is 500 page templates with accumulated styles and images in template. Since template storage is DB I think it is reasonable to allow mediumtext type for texts

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ node_modules
 .vscode
 /dbfiles
 /mkdocs_dist
+registry/lde/dbfiles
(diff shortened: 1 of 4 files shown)


b1ff:
@blackrabbit99 please describe in PR not only outcome of the change, but also a reason why this change is needed.

blackrabbit99:
> @blackrabbit99 please describe in PR not only outcome of the change, but also a reason why this change is needed.

thank you for mention. I have updated PR's description.

PR #390, opened 2022-04-04
docs: fully revised legacy umd bundles doc

bandantonio:


docs/umd_bundles_compatibility.md
```diff
@@ -1,20 +1,14 @@
-# Compatibility with legacy UMD bundles
+When adding ILC to the legacy website, you may have an issue when the code that is loaded via regular `<script>` tags works incorrectly.
```
b1ff, line 1: I would add header to a doc

b1ff (changes requested):

docs/umd_bundles_compatibility.md
bandantonio: @b1ff Do you want this heading to be in the markdown itself for the sake of completeness, or to be shown in the documentation?

In terms of the documentation, this heading is useless because MkDocs renders the proper heading anyway:

heading

yehor-manzhula (approved):

PR #395, opened 2022-04-07
Feature/update nodejs16

b1ff:


registry/server/app.ts
```diff
@@ -25,7 +25,7 @@ export default async (withAuth: boolean = true) => {
 
     app.get('/ping', pong);
 
-    app.use('/', serveStatic('client/dist'));
+    app.use('/', serveStatic('client/dist') as any);
```
sergeydevhub, line 28: Is "as any" used as a temporary solution?

sergeydevhub (approved):

registry/server/app.ts
b1ff: nope, types are not compatible unfortunately

nightnei (approved):

PR #400, opened 2022-04-12
docs: fully revised Parcels document

bandantonio:


yehor-manzhula (approved):

docs/parcels.md
```diff
@@ -2,188 +2,205 @@
 ignore_macros: true
 ---
 
-# Parcels
+## Overview
 
-> Parcels are an advanced feature of ILC / single-spa. We recommend that you use _Applications_ as the primary type of micro-frontend
```
b1ff, line 7: why did you decide to remove this note?

bandantonio: It sounds weird right at the beginning. You gave no overview to the user on what parcels are all about, but mentioned that you should probably ignore this feature. What's the point?

Parcels are located under the Advanced features in the docs sidebar, should we stress it again?
Furthermore, I can't see a meaningful connection between the complexity (or applicability) of the feature and our recommendation to use Applications as the primary type of the micro-frontend. These are more like two separate phrases rather than a coherent sentence.

For the note, I believe, we can leave just the recommendation closer to the end of the overview. As I said before, at the beginning of the page, it looks weird.

b1ff: ok

b1ff (approved):

PR #407, opened 2022-04-20
fix: correct behaviour of Header setCookie=ilc-i18n

nightnei:
1) previously clinet.js (ILC bundle) could set header (set-cookie=i18n). it caused unpredictable bugs. 
2) update of TailorX fixes the bug with overriding ILC's Headers by fragments Headers https://github.com/StyleT/tailorx/commit/2d83fd5fb89666ee3e0c93022e5a1a54e0333d9f

### DIFF
--- a/ilc/server/i18n.js
+++ b/ilc/server/i18n.js
@@ -4,8 +4,14 @@ const {intlSchema} = require('ilc-sdk/dist/server/IlcProtocol'); // "Private" im
 
 const i18nCookie = require('../common/i18nCookie');
 
+const isStaticFile = url => {
+    const extensions = ['.js', '.js.map'];
+
+    return extensions.some(ext => url.endsWith(ext));
+};
+
 const onRequestFactory = (i18nConfig, i18nParamsDetectionPlugin) => async (req, reply) => {
-    if (!i18nConfig.enabled || req.raw.url === '/ping') {
+    if (!i18nConfig.enabled || req.raw.url === '/ping' || isStaticFile(req.raw.url)) {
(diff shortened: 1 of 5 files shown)


yehor-manzhula (approved):

ilc/server/i18n.js
```diff
@@ -4,8 +4,14 @@ const {intlSchema} = require('ilc-sdk/dist/server/IlcProtocol'); // "Private" im
...
 
+const isStaticFile = url => {
+    const extensions = ['.js', '.js.map'];
+
+    return extensions.some(ext => url.endsWith(ext));
+};
+
 const onRequestFactory = (i18nConfig, i18nParamsDetectionPlugin) => async (req, reply) => {
-    if (!i18nConfig.enabled || req.raw.url === '/ping') {
+    if (!i18nConfig.enabled || req.raw.url === '/ping' || isStaticFile(req.raw.url)) {
```
yehor-manzhula, line 14: Does req.raw.url includes query parameters? In case if yes the condition might not work.

nightnei: it includes, but the condition is just to reject `.js` files e.g. http://localhost:8233/_ilc/client.js

yehor-manzhula: if for some reason you'll get http://localhost:8233/_ilc/client.js?q=1412341 condition won't work.

nightnei: done https://github.com/namecheap/ilc/pull/407/commits/38331a0d3e3fe7577a06a3a6d81c68c8fbae007c

PR #409, opened 2022-04-27
feat: test route for unsupported browser

nightnei:


ilc/server/app.js
```diff
@@ -82,6 +82,15 @@ module.exports = (registryService, pluginManager) => {
     // Route to test 500 page appearance
     app.get('/_ilc/500', async () => { throw new Error('500 page test error') });
 
+    // Route to test page for unsupported browser
+    app.get('/_ilc/unsupported_browser', async (req, res) => {
```
b1ff, line 86: why do we need it?

nightnei: At the current moment "our internal team" need to just develop the page and make different internal demos, so to provide this opportunity - I have created this route, for test purposes.
And then, when we decide how to handle unsupported browsers - we will create a complete solution for this feature, but anyway I think that this route will be useful for test purposes in future for sure.

b1ff: there is a lot of cons for this approach, i.e. sticking to a specific naming in template. While this is done for 500 it was a huge mistake per my opinion and we should not follow the same again.. Having said that and the fact that unsupported browsers page template might have any name and that it might be different pages for different domains and that this page is fully static template that is easy to setup locally and bind to usual route I quite do not understand need of this here.

PR #411, opened 2022-04-29
feat: optional slots for routes and templates

nightnei:


registry/server/appRoutes/routes/common.ts
```diff
@@ -0,0 +1,20 @@
+import db from '../../db';
```
b1ff, line 1: `common` is too generic name for a file, please name it `routesRepository`

b1ff (changes requested):

registry/server/appRoutes/routes/common.ts
nightnei: done

b1ff (approved):

PR #417, opened 2022-05-12
Update badge for demo website

bandantonio:
Because it shouldn't be red

badge


nightnei:
Use please "conventional commits"

nightnei:
Do you have some arguments regarding storing this image outside of the repo, instead of downloading it and put inside the repo?
Can we trust badgen.net?

bandantonio:
> Do you have some arguments regarding storing this image outside of the repo, instead of downloading it and put inside the repo?

Well, I don't. Just thought it would be a bit easier to maintain it. Anyway, svg as an asset looks fine too.

nightnei (approved):

PR #419, opened 2022-05-17
fix(chore): Added domainName restriction on getTemplateRendered api call

yehor-manzhula:


b1ff (approved):

ilc/server/errorHandler/ErrorHandler.spec.js
```diff
@@ -30,7 +36,7 @@ describe('ErrorHandler', () => {
 
     it('should show 500 error page with an error id', async () => {
         nock(config.get('registry').address).get('/api/v1/router_domains').reply(200, []);
-        nock(config.get('registry').address).get(`/api/v1/template/500/rendered?locale=en-US`).reply(200, {
+        nock(config.get('registry').address).get(`/api/v1/template/500/rendered?locale=en-US&domain=${address}`).reply(200, {
```
nightnei, line 39: you updated existed test case with an additional query `domain`. but should we create another test case to have different tests cases such as `with domain query` and `without domain query`?

registry/server/templates/routes/getRenderedTemplate.ts
```diff
@@ -28,13 +29,30 @@ const validateRequestBeforeGetTemplateRendered = validateRequestFactory([{
...
+                .join('routes', 'templates.name', 'routes.templateName')
+                .where({
+                    domainId: domainItem.id,
+                    name: templateName,
+                });
+        }
+    }
 
-    const [defaultTemplate] = await db.select().from<Template>('templates').where('name', templateName);
+    if (!defaultTemplate) {
```
nightnei, line 53: at line 57 we have the same condition and it's confusing. What do you think about creating different variables, e.g. `templateByDomain` and `defaultTemplate`?
or at least somehow rewrite it, since it's really hard to read when you see the code the first time.

nightnei (changes requested):

nightnei (approved):

PR #423, opened 2022-06-08
fix: dont create history item in registry if provided no changes

nightnei:


registry/server/versioning/services/Versioning.ts
```diff
@@ -53,11 +53,18 @@ export class Versioning {
                 throw new errors.VersioningError({m: `Unable to determine changeset for entity type "${conf.type}" & ID "${conf.id}"`});
             }
 
+            const data = currentData === null ? null : JSON.stringify(currentData);
+            const data_after = newData === null ? null : JSON.stringify(newData);
+
+            if (data === data_after) {
```
b1ff, line 59: might be still not valid comparison, because 
image

b1ff (changes requested):

registry/server/versioning/services/Versioning.ts
nightnei: you are right, but data is received from DB, that's why the order is always the same. and thats why I decided to not complicate the codebase and comparison process.
and anyway, this condition will be used in rare cases, and if condition is used then probably the are problems on the consumer side.

b1ff (approved):

PR #437, opened 2022-07-20
feat: add ability to customize static error page.

b1ff:
Now customer can specify its own error page if default is not suitable. Also default error page now is a valid HTML and more pretty than just text.


ilc/server/errorHandler/ErrorHandler.js
```diff
@@ -80,11 +83,34 @@ module.exports = class ErrorHandler {
...
+
+    #readStaticErrorPage(defaultContent) {
+        if (this.#staticFileContent != null) {
+            return this.#staticFileContent;
+        }
+
+        const staticFilePath = config.get('staticError.disasterFileContentPath');
+        try {
+            if (staticFilePath != null) {
+                this.#staticFileContent = readFileSync(staticFilePath).toString();
```
blackrabbit99, line 106: why sync ?
Maybe it makes sence serve it from memory ?

blackrabbit99 (changes requested):

ilc/server/errorHandler/ErrorHandler.js
b1ff: that is an idea. It is put into memory on first read

blackrabbit99: thank you

b1ff: you are welcome, can you approve PR?)

blackrabbit99 (approved):

PR #446, opened 2022-08-16
chore: pass localisation manifest URL in app config

blackrabbit99:
Application can optionally have link to localization manifest. So it can be used for example for handle immutable cache when your locale is dynamically updated


registry/client/src/dataProvider.js
```diff
@@ -43,7 +43,9 @@ const myDataProvider = {
         params.id = encodeURIComponent(params.id);
 
         return dataProvider.getOne(resource, params).then(v => {
+            console.log(v);
```
b1ff, line 46: seems like leftovers

registry/server/apps/routes/getApp.ts
```diff
@@ -28,9 +28,13 @@ const getApp = async (req: Request<GetAppRequestParams>, res: Response): Promise
 
     const [app] = await db.select().from<App>('apps').where('name', appName);
 
+    console.log('app = ');
+    console.log(app);
```
b1ff, line 32: can be removed

registry/server/apps/routes/getApp.ts
```diff
@@ -28,9 +28,13 @@ const getApp = async (req: Request<GetAppRequestParams>, res: Response): Promise
...
+
     if (!app) {
         res.status(404).send('Not found');
     } else {
+        console.log(preProcessResponse(app));
```
b1ff, line 37: also not needed I believe

registry/server/apps/routes/getApp.ts
blackrabbit99: looks like fixed locally but w/o push :(

yehor-manzhula (approved):

b1ff (approved):

registry/server/apps/routes/getApp.ts
blackrabbit99: solved

registry/server/apps/routes/getApp.ts
blackrabbit99: fixed

PR #453, opened 2022-09-26
fix(client): Added cors error to handle case when error event doesn't…

yehor-manzhula:
Handle cors error during runtime error

### DIFF
--- a/ilc/client/Client.js
+++ b/ilc/client/Client.js
@@ -12,7 +12,7 @@ import {
     addNavigationHook,
 } from './navigationEvents/setupEvents';
 
-import errors from './errors';
+import { CorsError } from './errors';
 
 import { triggerAppChange } from './navigationEvents';
 import navigationErrors from './navigationEvents/errors';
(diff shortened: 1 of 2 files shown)


ilc/client/errors/index.js
```diff
@@ -0,0 +1,7 @@
+import extendError from '@namecheap/error-extender';
+
+const errors = {};
+
+errors.CorsError = extendError('CorsError');
+
+export default Object.freeze(errors);
```
blackrabbit99, line 7: why default export and object.freeze ?

ilc/client/Client.js
```diff
@@ -173,6 +177,12 @@ export class Client {
             };
         }
 
+        if (!error && lineNo === 0 && colNo === 0) {
```
blackrabbit99, line 180: could you pls explain this condition ?

ilc/client/errors/index.js
yehor-manzhula: default export because it should be list of common client errors.
Object.freeze - to prevent list modification

blackrabbit99 (approved):

blackrabbit99 (approved):

PR #454, opened 2022-09-29
fix: fixed doubled transition manager when wrapped app is created.

b1ff:
Doubled TransitionManager led to multiple events about page completion which led to a different non-trivial bugs. TransitionManager is now possible to create multiple times, but only when previous instance removes event subscriptions.

### DIFF
--- a/ilc/client/WrapApp.js
+++ b/ilc/client/WrapApp.js
@@ -1,5 +1,5 @@
 import {flattenFnArray} from './utils';
-import transitionManager, {slotWillBe} from './TransitionManager/TransitionManager';
+import {slotWillBe} from './TransitionManager/TransitionManager';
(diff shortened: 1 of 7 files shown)


ilc/client/WrapApp.js
```diff
@@ -18,9 +18,9 @@ export default class WrapApp {
     #appState = 0;
     #wrapperState = 0;
 
-    constructor(wrapperConf, ssrOverrideProps) {
+    constructor(wrapperConf, ssrOverrideProps, transitionManager) {
         this.#wrapperConf = wrapperConf;
-        this.#transitionManager = transitionManager();
+        this.#transitionManager = transitionManager;
```
blackrabbit99, line 23: how is it used ? I can not see any invocation of private var

yehor-manzhula (approved):

PR #457, opened 2022-10-05
fix(registry): Added crossorigin attribute to scripts included by include directive

yehor-manzhula:


blackrabbit99 (approved):

b1ff:
I would not go this way, since it might break consumers, who might not have access-control-allow-origin on their CDNs

blackrabbit99:
> I would not go this way, since it might break consumers, who might not have access-control-allow-origin on their CDNs

Do you mean it can be breaking change ?
Can we set attribute by API ?

PR #459, opened 2022-10-07
Add localisation to shared libs

blackrabbit99:
* Shared Libs contains l10nManifest
* API for shared lib extended
* Client side API for read Shared Lib config

### DIFF
--- a/registry/server/routes/config.ts
+++ b/registry/server/routes/config.ts
@@ -121,8 +121,6 @@ router.get('/', async (req, res) => {
         return acc;
     }, {});
 
-    console.log(data);
-
     return res.send(data);
 });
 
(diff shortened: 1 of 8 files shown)


registry/server/routes/config.ts
```diff
@@ -115,6 +116,13 @@ router.get('/', async (req, res) => {
         return acc;
     }, {});
 
+    data.dynamicLibs = sharedLibs.reduce((acc, { name, spaBundle, l10nManifest }) => {
+        acc[name] = { spaBundle, l10nManifest };
+        return acc;
+    }, {});
+
+    console.log(data);
```
yehor-manzhula, line 124: looks like should be removed

blackrabbit99: Thanks. Good catch. Fixed

PR #466, opened 2022-10-19
fix(errors): Hardcode error names instead of using constructor.name a…

yehor-manzhula:
Hardcode error names instead of using constructor.name and rely on minifier options

### DIFF
--- a/ilc/client/errors/BaseError.spec.js
+++ b/ilc/client/errors/BaseError.spec.js
@@ -3,6 +3,27 @@ import { BaseError } from './BaseError';
 
 describe('BaseError', () => {
 
+    describe('extend', () => {
+        it('should return child class', () => {
+            const Child = BaseError.extend('ChildError');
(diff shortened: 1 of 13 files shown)


b1ff (approved):

ilc/client/errors/BaseError.spec.js
```diff
@@ -3,6 +3,27 @@ import { BaseError } from './BaseError';
...
+    describe('extend', () => {
+        it('should return child class', () => {
+            const Child = BaseError.extends('ChildError');
+            expect(new Child()).to.be.instanceOf(Child);
+            expect(new Child()).to.be.instanceOf(BaseError);
+        });
+  
+        it('should set correct child name', () => {
+            const Child = BaseError.extends('ChildError');
+            expect(new Child().name).to.be.instanceOf('ChildError');
```
b1ff, line 15: just `to.be.equal('ChildError')` should be enough

b1ff (approved):

PR #471, opened 2022-10-28
Feature/make ilc startup robust

blackrabbit99:


yehor-manzhula (approved):

ilc/client/ClientRouter.js
```diff
@@ -76,7 +76,10 @@ export default class ClientRouter extends EventEmitter {
     navigateToUrl = (url) => this.#singleSpa.navigateToUrl(this.#i18n.localizeUrl(url));
 
     getRelevantAppKind(appName, slotName) {
-        const appKind = this.#registryConf.apps[appName].kind;
```
b1ff, line 79: what if there is no app registered with `appName`?

ilc/common/Slot/SpaSlot.js
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * @constructor
+     * @param {{slotName: string, appName: string, appId: string}} slotName
+     * @param {IlcConfigRoot} rootConfig
+     */
+    constructor({applicationId, applicationName, slotName}, configRoot) {
+
+        assert(applicationId !== 'string', 'SpaSlot instance can not be initiated without applicationId');
+        assert(applicationName !== 'string', 'SpaSlot instance can not be initiated without applicationName');
+        assert(slotName !== 'string', 'SpaSlot instance can not be initiated without slotName');
```
b1ff, line 21: Those messages do not have all possible needed information, I would print in every message parameters that were specified, otherwise will be hard to debug

ilc/common/Slot/SpaSlot.js
```diff
@@ -0,0 +1,74 @@
...
+        this.#applicationName = applicationName;
+        this.#slotName = slotName;
+        this.#configRoot = configRoot;
+    }
+
+    /**
+     * @method
+     * @return {string} ApplicationName
+     */
+    getApplicationName() {
```
b1ff, line 33: can we use getters for it?

b1ff (changes requested):

b1ff:
It looks like PR handles invalid slot configuration, but original issue was about incorrect route configuration, i.e. app is forced to domain and do not exist in registry, but route that has that app is there.

ilc/common/Slot/SlotCollection.spec.js
```diff
@@ -0,0 +1,16 @@
...
+
+describe('SlotCollection', () => {
+    it('should init valid slot collection', () => {
+        const slotCollection = new SlotCollection(slotRawValid, registry);
+        expect(slotCollection.isValid()).to.be.true;
+    });
+
+    it('should init not valid slot collection', () => {
+        const slotCollection = new SlotCollection(slotRawNotValid, registry);
+        expect(() => {slotCollection.isValid()}).to.throw;
```
b1ff, line 13: it means that site is down if collection is invalid

ilc/client/ClientRouter.js
blackrabbit99: In that case verification will be invoked earlier and app will be failed with 500 error with explanation which app is not exist.

b1ff: Since method is public there is no guarantee that someone will not call it without validation, so I would rather accept `appKind` as method parameter or add validation here as well

blackrabbit99:
> It looks like PR handles invalid slot configuration, but original issue was about incorrect route configuration, i.e. app is forced to domain and do not exist in registry, but route that has that app is there.
Routes do not know nothing application. Route is assigned with template that contains slot and slot has a binding to apps. So that's why I think that the correct way to handle initial issue is to correctly process slots and verify validity of application attached to them

ilc/common/Slot/SpaSlot.js
blackrabbit99: We can. But I decided not to use it based on current code style. I would propose to switch to getter with transition to TS and setuping eslint into the project.

ilc/common/Slot/SpaSlot.js
b1ff: I mean, Slot could be created for any app and route, right? By throwing error that something is defined without something does not give us understanding what app has error. So I would print initial object that passed in ctor to error to have additional information

ilc/common/Slot/SlotCollection.js
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    /**
+     * @method isValid
+     * @return {boolean}
+     */
+    isValid() {
+        this.#slotCollection.forEach((slot) => {
+            if(!slot.isValid()) {
+                throw new Error(`Can not find application - ${slot.getApplicationName()}`);
```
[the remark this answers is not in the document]
blackrabbit99: We have only client specified error for now. So since we do not have them on backend lets go with a raw one.

b1ff (approved):

PR #474, opened 2022-11-03
fix(client): Rename error attribute from appName to name as it conflicts with appName provided by newrelic

yehor-manzhula:


ilc/client/Client.js
```diff
@@ -165,7 +165,7 @@ export class Client {
                 message: error.message,
                 data: {
                     ...errorInfo,
-                    appName,
```
b1ff, line 168: since it is an export to plugins it is a breaking change

b1ff (approved):

PR #475, opened 2022-11-07
Feature/add entry patch endpoint

blackrabbit99:
add `/api/v1/entries/:fqrn` endpoint.
It allows to patch fully qualified resource names (SharedLibs and Applications).
Motivation for it is providing single resource to update common properties for general entities.
Use case is patching localisation manifest in single way for SharedLibs and Applications


registry/server/common/services/assets/AssetsDiscoveryProcessor.ts
```diff
@@ -0,0 +1,44 @@
+import {
+    AssetsManifestReader,
+    AssetsManifest
+} from './AssetsManifestReader';
+
+export class AssetsDiscoveryProcessor {
```
b1ff, line 6: `Processor` looks not necessary, just `AssetsDiscovery` should be fine.

registry/server/common/services/assets/AssetsManifestReader.ts
```diff
@@ -0,0 +1,49 @@
...
+
+        try {
+            const response = await axios.get(assetsManifestUrl, {
+                responseType: 'json',
+            });
+
+            assetsManifestContent = response.data;
+
+        } catch (error) {
+            throw new AssetsManifestError(`"assetsDiscoveryUrl" ${assetsManifestUrl} is not available. Check the url via browser manually.`);
```
b1ff, line 30: looks like it hides original error. Can we propagate it?

registry/server/common/services/assets/AssetsManifestReader.ts
```diff
@@ -0,0 +1,49 @@
...
+        const assetsManifestContentValidator = Joi.object<AssetsManifest>({
+            spaBundle: Joi.string().required(),
+            cssBundle: Joi.string().optional(),
+            dependencies: Joi.object().pattern(Joi.string(), Joi.string()).optional(),
+        });
+
+        try {
+            return await assetsManifestContentValidator.validateAsync(assetsManifestContent);
+        } catch(error) {
+            throw new AssetsManifestError('"spaBundle" must be specified in the manifest file from provided "assetsDiscoveryUrl" if it was not specified manually');
```
b1ff, line 46: I'd propagate original Joi error

registry/server/common/services/entries/ApplicationEntry.ts
```diff
@@ -0,0 +1,98 @@
...
+        const  appManifest = await this.getManifest(appDTO.assetsDiscoveryUrl);
+
+        const appEntity = {
+            ...appDTO,
+            ...appManifest,
+        };
+
+        await db.versioning(user, {type: 'apps', id: appEntity.name}, async (trx) => {
+            await db('apps')
+                .insert(stringifyJSON(['dependencies', 'props', 'ssrProps', 'ssr', 'configSelector', 'discoveryMetadata'], appEntity))
```
b1ff, line 72: seems like a repetition from the `patch` method. I'd move to some kind of const.

registry/server/common/services/entries/ApplicationEntry.ts
```diff
@@ -0,0 +1,98 @@
...
+
+        const appEntity = {
+            ...appDTO,
+            ...appManifest,
+        };
+
+        await db.versioning(user, {type: this.entityName, id: this.identifier}, async (trx) => {
+            await db(this.entityName)
+                .where({ name: this.identifier })
+                .update(stringifyJSON([
```
b1ff, line 46: will this field cover l10n path tempalte?

b1ff (changes requested):
Would be nice to run auto-formatting on the changeset.

registry/server/common/services/entries/ApplicationEntry.ts
blackrabbit99: Yeap. This method travers keys and convert its value to strings so all fields that does not require stringification does not change own type

registry/server/common/services/assets/AssetsDiscoveryProcessor.ts
blackrabbit99: I 100% agree. But we have already `AssetsDiscovery` https://github.com/namecheap/ilc/blob/master/registry/server/common/services/AssetsDiscovery.ts.

So I have added Processor suffix because it's responsibility to form URL from files from assets discovery

registry/server/common/services/entries/ApplicationEntry.ts
blackrabbit99: Thank you. Updated

blackrabbit99:
> Would be nice to run auto-formatting on the changeset.

I have added prettier for registry. It is running on pre-commit hook.
In next PR will integrate it with ilc

registry/server/common/services/entries/EntryFactory.ts
```diff
@@ -0,0 +1,45 @@
...
+        APP: '@portal/',
+    };
+    public static getFqrnInstance(identifier: string) {
+        if (identifier.startsWith(this.resourceIdentifiers.SHARED_LIB)) {
+            const entityIdentifier = this.getEntityIdentifier(identifier, this.resourceIdentifiers.SHARED_LIB);
+            return this.getSharedLibInstance(entityIdentifier);
+        } else if (identifier.startsWith(this.resourceIdentifiers.APP)) {
+            const entityIdentifier = this.getEntityIdentifier(identifier, this.resourceIdentifiers.APP);
+            return this.getAppInstance(entityIdentifier);
+        } else {
```
yehor-manzhula, line 18: no need to have else condition here

registry/server/common/services/entries/error/FqrnError.ts
```diff
@@ -0,0 +1,3 @@
+export class FqrnError extends Error {
```
yehor-manzhula, line 1: Maybe it is better to rename to FullyQualifiedResourceNameError ?

registry/server/common/services/assets/AssetsManifestReader.ts
blackrabbit99: I think in that case we should not distinguish an error. Any network error shows us we have to debug accessibility of asset

registry/server/entries/entries.ts
```diff
@@ -0,0 +1,47 @@
...
+            return response.status(error.code).send(error.message);
+        }
+        throw error;
+    }
+
+    let results;
+    try {
+        results = await entryService.patch(params, { user: request.user });
+    } catch (error) {
+        if (error instanceof NotFoundFqrnError) {
```
yehor-manzhula, line 31: if (error instanceof NotFoundFqrnError || error instanceof ValidationFqrnError) {
    return response.status(error.code).send(error.message);
}

if (error instanceof Joi.ValidationError) {
    return response.status(422).send(joiErrorToResponse(error));
}
        
if (error instanceof AssetsManifestError) {
    return response.status(error.code).send(error.message);
}

registry/server/sharedLibs/routes/updateSharedLib.ts
```diff
@@ -26,34 +23,34 @@ const validateRequestBeforeUpdateSharedLib = validateRequestFactory([
...
-    await db.versioning(req.user, { type: 'shared_libs', id: sharedLibName }, async (trx) => {
-        await db('shared_libs').where({ name: sharedLibName }).update(sharedLib).transacting(trx);
-    });
+    let results;
 
-    const [updatedSharedLib] = await db.select().from<SharedLib>('shared_libs').where('name', sharedLibName);
+    try {
+        results = await sharedLibEntry.patch(sharedLib, { user: req.user });
+    } catch (error) {
+        if (error instanceof NotFoundFqrnError) {
```
yehor-manzhula, line 41: no need in else if conditions here

registry/server/entries/entries.ts
blackrabbit99: I think it make sence to optimize catching approach a bit later since all routes are supported single approach. 
In ideal case all errors should be presented in single format that could be parsed in the same manner.

registry/server/common/services/entries/EntryFactory.ts
blackrabbit99: Thanks. Changed.

registry/server/common/services/assets/AssetsManifestReader.ts
b1ff: I understand that, but if we do not have original error in error that is thrown and do not have it in logs, it would be hard to debug

yehor-manzhula (approved):

PR #481, opened 2022-12-07
chore: add info log to observer cache behaviour

blackrabbit99:


ilc/common/wrapWithCache.js
```diff
@@ -36,11 +52,24 @@ const wrapWithCache = (localStorage, logger, createHash = hashFn) => (fn, cacheP
...
+                    now,
+                    cachedAt: parsedItem ? parsedItem.cachedAt : undefined
+                }, logMessage('Item read from API. Invalidation finished.'));
+
                 return JSON.parse(localStorage.getItem(hash));
             }).catch(err => {
                 delete cacheResolutionPromise[hash];
 
                 if (localStorage.getItem(hash) === null) {
+                    logger.info({
```
b1ff, line 67: I'd use error here

b1ff (approved):

ilc/common/wrapWithCache.js
blackrabbit99: I also think in that way but want to analyze log on raw examples. There are some places in this logic I disagree with.

PR #484, opened 2022-12-14
chore: put request id to caching logs and log file checksum

blackrabbit99:
* update log information
* add config filecheck sum to understand when cache is stale on node
* pass request id to logs

### DIFF
--- a/ilc/common/wrapWithCache.js
+++ b/ilc/common/wrapWithCache.js
@@ -1,11 +1,10 @@
 const extendError = require('@namecheap/error-extender');
-
 const errors = {};
 errors.WrapWithCacheError = extendError('WrapWithCacheError');
 
 const wrapWithCache =
     (localStorage, logger, createHash = hashFn) =>
-    (fn, cacheParams) => {
+    (fn, cacheParams, contextStore) => {
         const { cacheForSeconds = 60, name } = cacheParams;
 
         if (typeof name !== 'string' || name.length === 0) {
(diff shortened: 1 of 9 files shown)


ilc/server/app.js
```diff
@@ -115,6 +132,17 @@ module.exports = (registryService, pluginManager) => {
             return;
         }
 
+        // temporary 20ms overhead
+        const hex = buildHashSum(JSON.stringify(JSON.stringify(registryConfig)));
```
b1ff, line 136: is double `stringify` intentional?

ilc/common/wrapWithCache.js
```diff
@@ -63,17 +64,16 @@ const wrapWithCache =
...
                             }),
                         );
-
                         logger.info(
                             {
                                 memoName: name,
                                 hash,
                                 now,
                                 cachedAt: parsedItem ? parsedItem.cachedAt : undefined,
+                                id: contextStore && contextStore.get('reqId'),
```
b1ff, line 73: do you have an example of result log?

ilc/server/app.js
blackrabbit99: NO. This is leftover. Fixing

blackrabbit99: Fixed

ilc/common/wrapWithCache.js
blackrabbit99: The case is `contextStore` is not available on client side.

b1ff: hmm, `id` format is strange. I thought that we can use it per our internal log specification, but with such format is impossible.
Also please note that having `id` in logs might not be transparent without context of this PR

b1ff (approved):

yehor-manzhula (approved):

yehor-manzhula (approved):

PR #488, opened 2022-12-22
feat(plugins): Added direct plugins import feature

yehor-manzhula:


b1ff (approved):

ilc/common/environment.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
```
blackrabbit99, line 1: can we use abstraction to read items from environment to not to put everything in object

ilc/server/pluginManager.js
```diff
@@ -2,18 +2,25 @@
 
 const path = require('path');
 const fg = require('fast-glob');
+const { LEGACY_PLUGINS_DISCOVERY } = require('../common/environment');
+const serverPlugins = require('../server.plugins');
```
blackrabbit99, line 6: avoid default export ?

ilc/client.plugins.js
```diff
@@ -0,0 +1 @@
+module.exports = [];
```
blackrabbit99, line 1: could you pls use abstraction because it is not clear why you decided to name file in this manner I mean with . separator

ilc/client/Client.js
```diff
@@ -122,6 +122,24 @@ export class Client {
         this.#configure();
     }
 
+    #loadPlugins() {
```
blackrabbit99, line 125: Not sure it is client responsibility

blackrabbit99 (changes requested):

ilc/common/Environment.js
```diff
@@ -0,0 +1,20 @@
...
+        this.#env = env;
+    }
+
+    isLegacyPluginsDiscoveryEnabled() {
+        return this.#env[this.#envVariableNames.legacyPluginsDiscovery] === 'false' ? false : true;
+    }
+}
+
+module.exports = {
+    Environment,
```
blackrabbit99, line 18: What is the idea behind exporting class and instance ?

blackrabbit99 (approved):

ilc/common/Environment.js
yehor-manzhula: Class export - for testing purposes
instance is binded to process.env to prevent code duplication

PR #501, opened 2023-03-20
Hreflang metatags

blackrabbit99:
* Add SSR hreflangs
* Support client side href langs

### DIFF
--- a/ilc/server/app.js
+++ b/ilc/server/app.js
@@ -160,9 +160,6 @@ module.exports = (registryService, pluginManager, context) => {
 
         const route = req.raw.router.getRoute();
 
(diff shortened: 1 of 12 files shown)


ilc/server/app.js
```diff
@@ -160,6 +160,9 @@ module.exports = (registryService, pluginManager, context) => {
 
         const route = req.raw.router.getRoute();
 
+        console.log('route = ', route);
```
yehor-manzhula: Should be replaced with logger or removed.

blackrabbit99: already done

yehor-manzhula (approved):

PR #504, opened 2023-03-21
feat: add canonical tag

blackrabbit99:
* Add SSR canonical TAG
* Add client side canonical handler

### DIFF
--- a/ilc/server/services/CanonicalTagService.js
+++ b/ilc/server/services/CanonicalTagService.js
@@ -1,13 +1,13 @@
 const IlcIntl = require('ilc-sdk/app').IlcIntl;
 const { context } = require('../context/context');
-const { removeQueryParams } = require('../../common/utils');
+const { removeQueryParams, addTrailingSlash } = require('../../common/utils');
 
 class CanonicalTagService {
     static getCanonicalTagForUrlAsHTML(url, locale, i18nConfig) {
         const store = context.getStore();
         const fullUrl = removeQueryParams(`${store.get('protocol')}://${store.get('domain')}${url}`);
         const localizedUrl = IlcIntl.localizeUrl(i18nConfig, fullUrl, { locale: locale || i18nConfig.default.locale });
-        return `<link rel="canonical" href="${localizedUrl}" data-ilc="1" />`;
+        return `<link rel="canonical" href="${addTrailingSlash(localizedUrl)}" data-ilc="1" />`;
     }
 }
 
--- a/ilc/server/tailor/configs-injector.spec.js
+++ b/ilc/server/tailor/configs-injector.spec.js
@@ -481,10 +481,10 @@ describe('configs injector', () => {
                             `</script>` +
                             `<script src="/_ilc/client.js" type="text/javascript"  async></script>` +
                             browserTimingHeader +
-                            `<link rel="alternate" hreflang="en-us" href="https://test.com/test/route" data-ilc="1">` +
-                            `<link rel="alternate" hreflang="ua-ua" href="https://test.com/ua/test/route" data-ilc="1">` +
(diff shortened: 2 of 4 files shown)


ilc/server/services/CanonicalTagService.js
```diff
@@ -0,0 +1,16 @@
+const IlcIntl = require('ilc-sdk/app').IlcIntl;
+const { context } = require('../context/context');
+const { removeQueryParams } = require('../../common/utils');
+
+class CanonicalTagService {
+    static getCanonicalTagForUrlAsHTML(url, locale, i18nConfig) {
+        const store = context.getStore();
+        const fullUrl = removeQueryParams(`${store.get('protocol')}://${store.get('domain')}${url}`);
```
b1ff, line 8: seems that might have XSS here, please double check that it is not possible through crafting malicious URL

ilc/server/tailor/configs-injector.spec.js
```diff
@@ -441,6 +484,7 @@ describe('configs injector', () => {
                             `<link rel="alternate" hreflang="en-us" href="https://test.com/test/route" data-ilc="1">` +
                             `<link rel="alternate" hreflang="ua-ua" href="https://test.com/ua/test/route" data-ilc="1">` +
                             `<link rel="alternate" hreflang="x-default" href="https://test.com/test/route" data-ilc="1">` +
+                            `<link rel="canonical" href="https://test.com/test/route" data-ilc="1" />` +
```
b1ff, line 487: we need to ensure that canonical always ends on `/`

b1ff: btw the same is probably applicable to the `hreflang` as well, because we now redirect customers from the URLs that are not ending on `/` the path that ends on `/` in order to protect from duplicating URLs in search engines. So I think it worth to generate both canonicals and hreflang with trailing slash

yehor-manzhula (approved):

ilc/client/CanonicalTagHandler.js
```diff
@@ -0,0 +1,33 @@
...
+
+    #handleRoutingChange = (event) => {
+        const url = removeQueryParams(event.target.location.href);
+        const canonicalTag = document.querySelector('link[rel="canonical"][data-ilc="1"]');
+        const localizedUrl = this.#i18n ? this.#i18n.localizeUrl(url) : url;
+
+        if (canonicalTag && canonicalTag.setAttribute) {
+            canonicalTag.setAttribute('href', localizedUrl);
+        } else {
+            this.#logger.error('CanonicalTagHandler: Can not find canonical tag on the page');
```
b1ff, line 30: Maybe in this case it worth to create one?

b1ff (changes requested):

ilc/server/services/CanonicalTagService.js
blackrabbit99: Interesting question. How I thought about it.
`protocol` is read from request state
`domain` as well
`url` is processed by i18n and passed after preliminary processing. 

Looking on other cases

blackrabbit99: We are using new URL interface under the hood so I expect internal sanitizing is enough to ensure URL is valid. If URL is valid it acts as safe URL.

ilc/client/CanonicalTagHandler.js
blackrabbit99: I decided not to create new one because absence of canonical on ssr is pretty critical so we need to react somehow on it. I expect stable behaviour here but in case of issues it is better to process it earlier because of SE impact could be crucial.

b1ff (approved):

b1ff (approved):

PR #516, opened 2023-05-29
Allow only get option requests

blackrabbit99:


ilc/server/app.js
```diff
@@ -34,9 +34,15 @@ module.exports = (registryService, pluginManager, context) => {
 
     app.addHook('onRequest', (req, reply, done) => {
         context.run({ request: req }, async () => {
-            const { url } = req.raw;
+            const { url, method } = req.raw;
             accessLogger.logRequest();
 
+            if (method !== 'GET' && method !== 'OPTION') {
```
b1ff, line 40: can we add tests pls?

blackrabbit99: added

PR #517, opened 2023-05-31
chore: not root subdomain wo own settings defined consumes global settings

blackrabbit99:


registry/tests/settings.spec.ts
```diff
@@ -263,6 +263,34 @@ describe(url, () => {
             }
         });
 
+        it('should return settings by domain name in case domain name is not root and it does not have own csp config', async () => {
```
b1ff, line 266: seems that case I've expected to see another one, something like

`should return default setting value if setting supports domain override but not overridden for passed domain`

b1ff (changes requested):

registry/tests/settings.spec.ts
blackrabbit99: Thank you. The description is more appropriate. Updated

b1ff (approved):

PR #518, opened 2023-06-01
chore: allow handle head requests

blackrabbit99:


ilc/server/app.js
```diff
@@ -37,8 +37,8 @@ module.exports = (registryService, pluginManager, context) => {
             const { url, method } = req.raw;
             accessLogger.logRequest();
 
-            if (method !== 'GET' && method !== 'OPTIONS') {
-                logger.warn(`Request method ${method} is not allowed`);
+            if (!['GET', 'OPTIONS', 'HEAD'].includes(method)) {
```
b1ff, line 40: I would also add that when `OPTIONS` and `HEAD` method should not trigger render of fragment, so below I'd changed to next

```js
     if (isStaticFile(url) || isHealthCheck(url) || ['OPTIONS', 'HEAD'].includes(method)) {
                return done();
            }
```

blackrabbit99: I've added this check to not to read settings on Options and Head requests

b1ff (approved):

PR #519, opened 2023-07-03
Feature/ssr only app

b1ff:


b1ff:
As for now it allows to create SSR only app, but not update. With updates and our usage in terraform like environment and with potentially other users of the registry API it is a little bit challenging to provide ability to change client side app to SSR only, so it should be processed through delete + create new.

yehor-manzhula (approved):

PR #529, opened 2023-08-21
Fix incorrect validation of the templates correctness

stas-nc:
Fix of template check in the ilc/server
Also added validation to the registry/server


ilc/server/tailor/configs-injector.js
```diff
@@ -26,14 +26,9 @@ module.exports = class ConfigsInjector {
...
 
-        if (
-            typeof document !== 'string' ||
-            !document.includes('<html') ||
-            !document.includes('</head>') ||
-            !document.includes('<head>') ||
-            !document.includes('</body>') ||
-            !document.includes('<body>')
-        ) {
+        const validHtmlRegex = /<html\s*.*?>.*<head\s*.*?>.*<\/head>.*<body\s*.*?>.*<\/body>.*<\/html>/s;
```
blackrabbit99, line 29: duplicated regexp

blackrabbit99 (changes requested):

ilc/server/tailor/configs-injector.js
stas-nc: It's a different app. Do we have here some place for shared constants/utils between apps?

ilc/server/registry/Registry.js
```diff
@@ -162,10 +163,14 @@ module.exports = class Registry {
             });
         }
 
-        let templateStr = res.data.content;
+        const rawTemplate = res.data.content;
 
-        let lastMatchOffset = templateStr.lastIndexOf('<ilc-slot');
-        templateStr = templateStr.replace(/<ilc-slot\s+id="(.+)"\s*\/?>/gm, function (match, id, offset) {
+        if (!isTemplateValid(rawTemplate)) {
+            throw new Error('Invalid template');
```
b1ff, line 169: Please add template name in the error

registry/server/templates/services/renderTemplate.ts
```diff
@@ -29,6 +30,10 @@ type RenderTemplateResult = {
 };
 
 async function renderTemplate(template: string): Promise<RenderTemplateResult> {
+    if (!isTemplateValid(template)) {
+        throw new Error('HTML template has invalid structure');
```
b1ff, line 34: please add template name to the error

b1ff (changes requested):

registry/server/templates/services/renderTemplate.ts
stas-nc: Here it is problematic as in some cases it is not possible to get the template name in the context of execution.
For instance, this method is used in the update template route, where the renderTemplate function is called in the validator, and the template name is not available there.

ilc/server/registry/Registry.js
stas-nc: added

registry/server/templates/services/renderTemplate.ts
b1ff: understood, could you please then add into that method log that template is going to be updated, with template details

stas-nc: I thought that we could actually catch and wrap this error, so it could log that the template "name" wasn't created/updated and the cause error is "Invalid html". 
What do you think?

b1ff: yeap, good option. Works as well

yehor-manzhula (approved):

ilc/server/registry/Registry.spec.js
```diff
@@ -279,5 +279,90 @@ describe('Registry', () => {
...
+        it('should throw an error when a document does not have <head>', async () => {
+            nock(address).get('/api/v1/template/anotherErrorTemplate/rendered').reply(200, {
+                content: '</head><body>Hi there! I do not have head tag.</body>',
+            });
+
+            const registry = new Registry(address, mockPreheat, logger);
+
+            await chai
+                .expect(registry.getTemplate('anotherErrorTemplate'))
+                .to.eventually.rejectedWith('Invalid structure  in template "anotherErrorTemplate"');
```
yehor-manzhula, line 304: Looks like extra-spaces

ilc/server/registry/Registry.js
```diff
@@ -162,10 +163,14 @@ module.exports = class Registry {
             });
         }
 
-        let templateStr = res.data.content;
+        const rawTemplate = res.data.content;
 
-        let lastMatchOffset = templateStr.lastIndexOf('<ilc-slot');
-        templateStr = templateStr.replace(/<ilc-slot\s+id="(.+)"\s*\/?>/gm, function (match, id, offset) {
+        if (!isTemplateValid(rawTemplate)) {
+            throw new errors.RegistryError({ message: `Invalid structure  in template "${templateName}"` });
```
yehor-manzhula, line 169: Extra spaces in error message

ilc/server/registry/Registry.spec.js
```diff
@@ -279,5 +279,90 @@ describe('Registry', () => {
...
+        it('should throw an error when a document does not have </body>', async () => {
+            nock(address).get('/api/v1/template/anotherErrorTemplate/rendered').reply(200, {
+                content: '<head></head><body>Hi there! I do not have closed body tag.',
+            });
+
+            const registry = new Registry(address, mockPreheat, logger);
+
+            await chai
+                .expect(registry.getTemplate('anotherErrorTemplate'))
+                .to.eventually.rejectedWith('Invalid structure  in template "anotherErrorTemplate"');
```
yehor-manzhula, line 352: Extra spaces

ilc/server/registry/Registry.spec.js
stas-nc: fixed in all places.

b1ff (approved):

blackrabbit99 (approved):

PR #531, opened 2023-09-01
feat(registry): add structured logs support

stas-nc:


registry/server/index.ts
```diff
@@ -1,16 +1,39 @@
 import path from 'path';
 
+['log', 'warn', 'error'].forEach(function (method) {
```
b1ff, line 3: this code looks a little bit not transparent to me. Could you please leave a comment why is it written in this way and explaining what does it do

registry/server/index.ts
```diff
@@ -1,16 +1,39 @@
...
 process.env.NODE_CONFIG_DIR = path.resolve(__dirname, '../config');
 require('newrelic'); //Should be lower than NODE_CONFIG_DIR env var definition
 
-import app from './app';
-import server from './server';
+let log = console.error;
 
 (async () => {
     try {
+        loadPlugins();
```
b1ff, line 31: I see that `loadPlugins()` is called 2 times here and inside `app()`. Is it done intentionally?

registry/server/util/logger.ts
```diff
@@ -0,0 +1,52 @@
+import { Logger } from 'ilc-plugins-sdk';
+import { getPluginManagerInstance } from './pluginManager';
+import { storage } from '../middleware/context';
+
+const hasKey = <T extends object>(obj: T, k: keyof any): k is keyof T => k in obj;
+
+const loggerProxyHandler: ProxyHandler<Logger> = {
+    get(target, prop) {
+        const requestIdKey = getPluginManagerInstance().getReportingPlugin().requestIdLogLabel ?? 'operationId';
```
b1ff, line 9: it seems that this one is needed only inside `if` body

registry/server/util/logger.ts
```diff
@@ -0,0 +1,52 @@
...
+        const origMethod = hasKey(target, prop) && target[prop];
+        if (typeof origMethod === 'function') {
```
b1ff, line 11: could you please invert condition to reduce nesting. And probably in order not to break to much, the `else` case should be handled as well

registry/server/util/pluginManager.ts
```diff
@@ -0,0 +1,17 @@
+import { PluginManager } from 'ilc-plugins-sdk';
+import manifest from '../../server.plugins.manifest.json';
+
+let pluginManagerInstance: PluginManager;
+
+export function loadPlugins() {
+    if (!pluginManagerInstance) {
+        pluginManagerInstance = new PluginManager(...manifest.manifest.plugins);
```
b1ff, line 8: is there any potential issue to initialize it directly in module? 
i.e. `const pluginManagerInstance = new PluginManager(...)`

b1ff (changes requested):

registry/server/util/pluginManager.ts
stas-nc: it will be triggered in the import ESM module phase, so it makes it harder to control flow.

registry/server/index.ts
stas-nc: fixed

b1ff (approved):

PR #538, opened 2023-10-12
Feature/node20

stas-nc:


yehor-manzhula (approved):

ilc/Dockerfile
b1ff, line 1: could you please update npm to v10 in docker image  here and in other places as well

b1ff (changes requested):

ilc/Dockerfile
stas-nc: Node v 20.8.0 comes with npm v10 installed
```
ilc git:(feature/node20) docker run --rm ilc npm -v
10.1.0
```

b1ff (approved):

PR #543, opened 2023-11-03
fix: when query parameters are changed on a page - ilc still tries to scroll to top.

b1ff:
Also fixed minor issues in scroll restoration work, like possibility to inject CSS selectors or raise errors by invalid URLs.


yehor-manzhula (approved):

ilc/client/TransitionManager/ScrollController/ScrollController.js
```diff
@@ -1,10 +1,18 @@
-export class UrlHashController {
+export class ScrollController {
     // @todo write e2e tests for hash position restoring
     #hashStoreNode = document.body;
     #hashStoreAttribute = 'ilcTempStoredHash';
+    #lastVisitedUrl = this.location.pathname;
+    #shouldScrollToTop = true;
+
+    get location() {
+        return window.location;
```
blackrabbit99, line 9: maybe it is better to use router. I would avoid calling `window.location` because scroll process happens during page transition by triggering route.change.

b1ff: yeah, I also didn't like it, it was spread over the class. I though about injecting something upfront, but most likely it would  increase PR changes, so I decided to keep improvement small and isolate window usage within one getter. Next step might be injection from the above

ilc/client/TransitionManager/ScrollController/ScrollController.js
```diff
@@ -15,26 +23,28 @@ export class UrlHashController {
         const node = this.#hashStoreNode;
         // @todo: looks like it never used so storing is useless
         node.removeAttribute(this.#hashStoreAttribute);
-        this.#scrollToHash();
+        this.#restoreScrollOnNavigation();
```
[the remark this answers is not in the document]
b1ff: but the method is called `restore` (try to expand code in github interface). Should be ok from my side

blackrabbit99 (approved):

stas-nc (approved):

PR #548, opened 2023-11-07
chore: resolve issue with unhandled promise when slot blocker timeout…

blackrabbit99:
… occurs


ilc/client/TransitionManager/TransitionManagerTimeout.spec.js
```diff
@@ -6,7 +6,7 @@ import ilcEvents from '../constants/ilcEvents';
 import { slotWillBe, TransitionManager } from './TransitionManager';
 import { CssTrackedApp } from '../CssTrackedApp';
 
-describe('TransitionManager Timeout', () => {
+describe.only('TransitionManager Timeout', () => {
```
yehor-manzhula, line 9: I beleive it is not expected and should be reverted

stas-nc (approved):

PR #551, opened 2023-11-08
fix(registry): fix openid identifier match (case-insensitive)

stas-nc:


registry/server/auth.ts
```diff
@@ -318,10 +318,14 @@ export default async (app: Express, settingsService: SettingsService, config: an
...
-        identifier,
-    });
+    const user = await db
+        .select()
+        .from('auth_entities')
+        .first('identifier', 'id', 'role', 'secret')
+        .where({
+            provider,
+        })
+        .andWhereRaw('LOWER(identifier) = LOWER(?)', [identifier]);
```
b1ff, line 328: won't it break mysql?
Also is it effective?

yehor-manzhula (approved):

registry/server/auth.ts
stas-nc: Tests on sqlite and mysql pass
Additional test for case-insensitive identifier added.

b1ff (approved):

registry/server/auth.ts
stas-nc: Regarding effectiveness https://stackoverflow.com/a/7005332/5825564
Can add an index on lower(identity column) though I'm not sure that this query and table is used very much.

b1ff: Ok, lets see how it will affect DB load

e2e/spec/news.spec.e2e.ts
```diff
@@ -12,7 +12,7 @@ Scenario('should open a news page and show news sources', async ({ I, newsPage:
     I.see('Pick a news source', newsPage.bannerHeadline);
 });
 
-Scenario.skip('should open an article page from a direct link', async ({ I, newsPage: newsPage }) => {
+Scenario.only('should open an article page from a direct link', async ({ I, newsPage: newsPage }) => {
```
yehor-manzhula, line 15: is only expected here?

stas-nc: sorry, not ready yet

yehor-manzhula (approved):

PR #558, opened 2023-11-17
Feat/add cache wrapper

yehor-manzhula:
I rewrote the wrapWithCache function into the CacheWrapper class to store progress promises inside the class. In case there is misuse of the wrap function, it will not result in two parallel requests.

### DIFF
--- a/ilc/server/registry/factory.js
+++ b/ilc/server/registry/factory.js
@@ -6,8 +6,4 @@ const reportingPluginManager = require('../plugins/reportingPlugin');
 const { context } = require('../context/context');
 const logger = reportingPluginManager.getLogger();
 
-module.exports = new Registry(
-    config.get('registry.address'),
-    new CacheWrapper(localStorage, logger, context.getStore()),
-    logger,
-);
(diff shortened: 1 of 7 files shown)


ilc/server/registry/factory.js
```diff
@@ -1,9 +1,13 @@
...
+const CacheWrapper = require('../../common/CacheWrapper');
 const reportingPluginManager = require('../plugins/reportingPlugin');
-
+const { context } = require('../context/context');
 const logger = reportingPluginManager.getLogger();
 
-module.exports = new Registry(config.get('registry.address'), wrapFetchWithCache(localStorage, logger), logger);
+module.exports = new Registry(
+    config.get('registry.address'),
+    new CacheWrapper(localStorage, logger, context.getStore()),
```
b1ff, line 11: I might be wrong but it seems that `getStore` has to be called inside the call stack, because it will dump async context, and if we do it on the module level, it always will be the same...

yehor-manzhula: Updated

b1ff (approved):

PR #559, opened 2023-11-20
feat(ilc/client): expose app configuration to loadApp  method

stas-nc:
Allow to skip css loading in loadApp method based on option `injectGlobalCss`.
`window.ILC.loadApp` loads css via `CssTrackedApp` (backwards compatible).
Add `cssBundle` to the result object of `loadApp` method.

### DIFF
--- a/ilc/client/BundleLoader.js
+++ b/ilc/client/BundleLoader.js
@@ -44,19 +44,33 @@ export class BundleLoader {
         }
     }
 
-    loadApp(appName) {
-        const app = this.#getApp(appName);
-        if (!app.spaBundle) {
+    /**
+     *
+     * @param {string} appName
+     * @param {boolean} options.injectGlobalCss to css with <link> element in <head>
+     * @returns Promise<object> application
+     */
+    loadApp(appName, { injectGlobalCss = true } = {}) {
+        const applicationConfig = this.#getApp(appName);
+        if (!applicationConfig.spaBundle) {
             // it is SSR only app
             return Promise.resolve(emptyClientApplication);
         }
 
         return this.#moduleLoader.import(appName).then((appBundle) => {
             const sdkInstanceFactory = this.#sdkFactoryBuilder.getSdkFactoryByApplicationName(appName);
-            const rawCallbacks = this.#getAppSpaCallbacks(appBundle, app.props, { sdkFactory: sdkInstanceFactory });
-            return typeof app.cssBundle === 'string'
-                ? new CssTrackedApp(rawCallbacks, app.cssBundle, this.#delayCssRemoval).getDecoratedApp()
-                : rawCallbacks;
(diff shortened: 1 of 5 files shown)


e2e/spec/missing-slot.spec.e2e.ts
```diff
@@ -37,6 +37,7 @@ Scenario('(CSR) should correctly open page where app requires missing slot in te
 
     const consoleMessages: ConsoleMessage[] = await I.grabBrowserLogs();
     const messagesText = consoleMessages.map(v => v.text());
+    console.log("🚀 ~ file: missing-slot.spec.e2e.ts:40 ~ Scenario ~ messagesText:", messagesText)
```
blackrabbit99, line 40: Leftover or we do need this comment in prod ?

ilc/client/BundleLoader.js
```diff
@@ -93,6 +107,16 @@ export class BundleLoader {
         return app;
     };
 
+    /**
+     * Merge config to to application object
+     * @param {object} app
+     * @param {object} config
+     * @returns {object}
+     */
+    #enrichApplicationWithConfig(app, applicationConfig) {
```
blackrabbit99, line 116: We usually read config from rootConfig object. Should not we do the same here ?

ilc/client/CssTrackedApp.js
```diff
@@ -26,6 +26,7 @@ export class CssTrackedApp {
             mount: this.mount,
             unmount: this.unmount,
             update: this.update,
+            __CSS_TRACKED_APP__: true,
```
blackrabbit99, line 29: Could you pls explain logic behind it ?

blackrabbit99 (commented):
We have to discuss this PR

ilc/client/CssTrackedApp.js
stas-nc: Property added for unit test and easier debugging.

e2e/spec/missing-slot.spec.e2e.ts
stas-nc: Debugging e2e test, already removed.

ilc/client/BundleLoader.js
stas-nc: Do you mean to expose an additional method to the window.ILC, instead of adding config to the app object?

blackrabbit99: no. window.ILC is exposed for consumers. From what I could see you do not need additional data in your case. Maybe I missed your context. Could you elaborate with motivation of that logic please ?

stas-nc: I need to access the application config on the consumer side (especially cssBundle) in order to allow custom css loading from the consumer side.

b1ff (approved):

yehor-manzhula (approved):

PR #560, opened 2023-11-20
Ilc typescript support

blackrabbit99:
ILC supports typescript v1


ilc/tsconfig.client.json
stas-nc, line 1: What about extending some base config in the root folder?

ilc/build/webpack.js
```diff
@@ -19,6 +19,16 @@ module.exports = {
     mode: 'production',
     module: {
         rules: [
+            {
+                test: /\.(js|ts)$/,
+                use: {
+                    loader: 'ts-loader',
```
b1ff, line 25: no any rust? :(

ilc/package.json
```diff
@@ -4,15 +4,16 @@
   "description": "",
   "main": "server/index.js",
   "scripts": {
-    "test": "cross-env NODE_ENV=test mocha",
+    "test": "npm run build:server && cross-env NODE_ENV=test mocha",
     "test:ci": "npm run test -- --forbid-only --exit",
     "test:watch": "npm run test -- --watch",
```
b1ff, line 9: I assume now this command won't work. Maybe better `ts-node`?

b1ff (changes requested):

ilc/package.json
blackrabbit99: this works. Idea is source require conf file like list of plugins so the easiest way to run mocha tests is to run build and execute tests from dist

ilc/tsconfig.client.json
blackrabbit99: That is good idea. Will add base config

ilc/package.json
b1ff: can you explain how it work in `--watch` mode, if I change typescript file and tests are re-run?

blackrabbit99: Watch mode in such setup won't work. I got your point.

b1ff (approved):

ilc/server/serveStatic.js
```diff
@@ -7,11 +7,11 @@ module.exports = function (isProduction) {
         return serveStatic('public');
     }
 
-    require('../systemjs/build');
+    require('../../systemjs/build');
```
yehor-manzhula, line 10: Why the path changed?

yehor-manzhula (approved):

PR #563, opened 2023-11-30
feat(ilc-client): pass application manifest to the ilc-sdk instance

stas-nc:


ilc/client/Sdk/SdkFactoryBuilder.js
```diff
@@ -20,6 +20,7 @@ export class SdkFactoryBuilder {
             i18n: {
                 manifestPath,
             },
+            manifest: appConfig,
```
b1ff, line 23: A little bit confused by this line. Manifest usually is understood by us as path to js and css. App config is app config that mostly user defined. Can you please correct some naming in the code so they are not intersected?

stas-nc: renamed

b1ff (approved):

blackrabbit99 (approved):

PR #581, opened 2024-03-15
feat: return client errors on internal ilc render template endpoint

stas-nc:
return NotFound and BadRequest errors in HTML on internal _ilc … render template endpoint
refactor some parts to ts
move default HTML page to html file
fix the issue with async context loss
add handleClientError() method


yehor-manzhula (approved):

ilc/server/app.js
```diff
@@ -23,17 +18,22 @@ const accessLogger = new AccessLogger(config);
...
     const logger = reportingPluginManager.getLogger();
 
     const app = fastify(appConfig);
 
+    const asyncResourceSymbol = Symbol('asyncResource');
+
     app.addHook('onRequest', (req, reply, done) => {
         context.run({ request: req }, async () => {
             try {
+                const asyncResource = new AsyncResource('fastify-request-context');
```
blackrabbit99, line 33: context.run is using Async Local Storage already maybe you could reuse it instead of running additional context ?

blackrabbit99 (changes requested):

ilc/server/app.js
stas-nc: There is an issue with losing context between async hooks.
See this code https://github.com/fastify/fastify-request-context/blob/master/index.js#L46

blackrabbit99 (approved):

PR #595, opened 2024-08-08
fix(ilc): remove polyfills

stas-nc:
Remove outdated polyfills as polyfill.io is not available.

### DIFF
--- a/ilc/package-lock.json
+++ b/ilc/package-lock.json
@@ -85,7 +85,6 @@
         "webpack": "^5.88.2",
         "webpack-cli": "^5.1.4",
         "webpack-dev-middleware": "^6.1.1",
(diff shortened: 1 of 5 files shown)


blackrabbit99:
> Remove outdated polyfills as polyfill.io is not available.

Could you please check it does not create breaking changes ?

yehor-manzhula (approved):

blackrabbit99:
> @blackrabbit99 Removed polyfills are supported by the majority of modern browsers, and our [Browser Support] document.

Our browser doc is internal not external here I believe some of consumers could support older browsers.

blackrabbit99 (changes requested):
I believe it is breaking change and could not be released without major version

blackrabbit99 (approved):

PR #597, opened 2024-08-14
fix: improve assets discovery worker stability.

b1ff:
It now does not allow to have parallel update process of the same assets. 
It ensures timeouts for all the async processes to avoid long or never ending requests to db or HTTPs. 
It ensures that even if one entity process is failed - others will be successfully updated.

### DIFF
--- a/registry/server/common/services/assets/AssetsDiscovery.ts
+++ b/registry/server/common/services/assets/AssetsDiscovery.ts
@@ -3,98 +3,147 @@ import urljoin from 'url-join';
 import deepEqual from 'deep-equal';
 
 import knex from '../../../db';
-import manifestProcessor from './assetsManifestProcessor';
+import manifestProcessor, { ManifestData } from './assetsManifestProcessor';
 import AssetsDiscoveryWhiteLists from './AssetsDiscoveryWhiteLists';
 import { getLogger } from '../../../util/logger';
 import { parseJSON } from '../json';
 import { axiosErrorTransformer } from '../../../util/axiosErrorTransformer';
+import { startBackgroundTransaction } from 'newrelic';
+
+type AssetsDiscoveryEntity = {
+    [key: string]: any;
+    assetsDiscoveryUrl: string;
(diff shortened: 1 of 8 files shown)


registry/server/common/services/assets/AssetsDiscovery.ts
```diff
@@ -3,98 +3,147 @@ import urljoin from 'url-join';
...
     }
 
     stop() {
-        clearInterval(this.timerId!);
+        clearTimeout(this.timerId!);
+    }
+
+    private async iteration() {
+        return startBackgroundTransaction(this.tableName, AssetsDiscovery.constructor.name, async () => {
+            const now = Math.floor(Date.now() / 1000);
```
yehor-manzhula, line 62: I still beleive that enire logic should be changed in order to take in a count
all the previous responses time/timeouts in updatedTime parameter.

yehor-manzhula (approved):

registry/server/common/services/assets/AssetsDiscovery.ts
b1ff: Noted

blackrabbit99 (approved):

stas-nc (approved):

registry/server/common/services/assets/AssetsDiscovery.ts
```diff
@@ -3,98 +3,147 @@ import urljoin from 'url-join';
...
         this.tableId = tableId;
         this.intervalSeconds = intervalSeconds;
     }
 
     start(delay: number = 1000) {
-        this.timerId = setInterval(
-            () =>
-                this.iteration().catch((err) => {
+        const runLoop = () => {
+            this.timerId = setTimeout(async () => {
```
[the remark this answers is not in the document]
b1ff: setInterval can run processes in paralell, because if process takes more time than interval - interval won't wait completion before next interval is started, that is why it can lead to sum resources for all parallel processes

PR #613, opened 2024-10-03
feat(registry): Add endpoint to only change a template's content

harnyk:
This PR introduces a new endpoint:

 - `PATCH /api/v1/template/:name`;
   - body: `{ content: string }`

The endpoint allows to update only template's content without deleting its localized versions

### DIFF
--- a/registry/server/templates/services/templatesRepository.ts
+++ b/registry/server/templates/services/templatesRepository.ts
@@ -154,7 +154,7 @@ export class TemplatesRepository {
             }
         });
 
-        const savedTemplate = await this.mustReadTemplateWithAllVersions(template.name);
+        const savedTemplate = await this.readTemplateWithAllVersionsOrFail(template.name);
 
         return savedTemplate;
     }
@@ -189,7 +189,7 @@ export class TemplatesRepository {
             },
         );
 
-        const updatedTemplate = await this.mustReadTemplateWithAllVersions(templateName);
+        const updatedTemplate = await this.readTemplateWithAllVersionsOrFail(templateName);
 
         return { type: 'ok', template: updatedTemplate };
     }
@@ -211,7 +211,7 @@ export class TemplatesRepository {
                 .transacting(trx);
         });
 
-        const updatedTemplate = await this.mustReadTemplate(templateName);
+        const updatedTemplate = await this.readTemplateOrFail(templateName);
 
         return { type: 'ok', template: updatedTemplate };
     }
@@ -261,7 +261,7 @@ export class TemplatesRepository {
         return { type: 'ok' };
     }
 
-    private async mustReadTemplate(templateName: string): Promise<VersionedRecord<Template>> {
+    private async readTemplateOrFail(templateName: string): Promise<VersionedRecord<Template>> {
         const [maybeTemplate] = await db
             .selectVersionedRowsFrom<Template>(Tables.Templates, 'name', EntityTypes.templates, [
                 `${Tables.Templates}.*`,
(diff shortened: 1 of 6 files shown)


registry/tests/templates.spec.ts
```diff
@@ -576,6 +576,49 @@ describe(`Tests ${example.url}`, () => {
...
+                .patch(example.url + example.correct.name)
+                .send({ content: example.updated.content })
+                .expect(200);
+            expect(response.body).deep.equal({
+                ...example.updated,
+                versionId: response.body.versionId,
+            });
+        });
+
+        it('should not delete localized versions of the template on partial update', async () => {
```
wRLSS, line 601: Maybe even explicitly notify api user about passing localisedVersions is not supported?

harnyk: like in which way?

wRLSS: validation in case locales were sent

harnyk: already implemented. Just added a test

registry/server/templates/services/templatesRepository.ts
```diff
@@ -182,6 +194,28 @@ export class TemplatesRepository {
         return { type: 'ok', template: updatedTemplate };
     }
 
+    async partialUpdateTemplate(
+        templateName: string,
+        payload: PartialUpdateTemplatePayload,
+        user: User | undefined,
+    ): Promise<PartialUpdateTemplateResult> {
+        const templateExists = await this.db(Tables.Templates).where({ name: templateName }).select(1).first();
```
wRLSS, line 202: doesnt `db` have some firm syntax for `findOne` ?

registry/server/templates/services/templatesRepository.ts
```diff
@@ -227,6 +261,19 @@ export class TemplatesRepository {
         return { type: 'ok' };
     }
 
+    private async mustReadTemplate(templateName: string): Promise<VersionedRecord<Template>> {
```
wRLSS, line 264: usual syntax for this in ORMs afaik is `get` || `findOrFail`, tbh it's slightly harder to read `mustReadTemplate`

registry/tests/templates.spec.ts
```diff
@@ -576,6 +576,57 @@ describe(`Tests ${example.url}`, () => {
         });
     });
 
+    describe('Partial Update', () => {
+        it("should not partial update any record if record doesn't exist", async () => {
```
wRLSS, line 580: (minor) should not partial update > should not partially update 
not so robotic

registry/tests/templates.spec.ts
```diff
@@ -576,6 +576,57 @@ describe(`Tests ${example.url}`, () => {
...
+            const incorrect = { name: 123 };
```
wRLSS, line 581: `incorrect` -> `nonExisting`, slightly better semantics on why it's incorrect, some might assume that's due to int inside name

registry/tests/templates.spec.ts
```diff
@@ -576,6 +576,57 @@ describe(`Tests ${example.url}`, () => {
...
+            await req
+                .patch(example.url + incorrect.name)
+                .send({ content: example.correct.content })
+                .expect(404, 'Not found');
+        });
+
+        it('should successfully partial update record', async () => {
```
wRLSS, line 588: please slightly less robotic description

stas-nc (approved):

registry/server/templates/services/templatesRepository.ts
harnyk: No, it doesn't. This is the optimal syntax for this kind of queries.

Why `first()`? Because I need either a row or `null`.
Why `select(1)`? Because I don't need to transfer any data at all.

wRLSS (approved):

PR #614, opened 2024-10-03
feat(ilc): disable newrelic for lde

m2broth:
Disable new relic on client for LDE by removing newrelic script from page in the LDE mode (ILC-overrideConfig cookie is set). 
Added custom tags <!- Prod only start -> and <!- Prod only end -> for template parts that will be removed in LDE mode


stas-nc (approved):

stas-nc (approved):

yehor-manzhula (approved):

ilc/server/tailor/configs-injector.js
```diff
@@ -246,4 +250,11 @@ module.exports = class ConfigsInjector {
         nrCode = nrCode.replace(/<script.*?>(.*)<\/script\s*>/s, '$1');
         return this.#nrCustomClientJsWrapper.replace('%CONTENT%', nrCode);
     };
+
+    #wrapWithRemoveBeforeParsing = (content) =>
+        `<!-- TailorX: Remove before parsing START -->${content}<!-- TailorX: Remove before parsing END -->`;
+
+    #wrapTagsToRemove(content, regex) {
```
b1ff, line 257: I cannot get why is it needed? Newrelic that is put here can be removed right away, these tags should be set directly in the templates if they supposed to be removed on LDE

b1ff (changes requested):

yehor-manzhula (approved):

ilc/server/tailor/configs-injector.js
m2broth: I tried to add a general logic for removing parts from templates and implement LDE check on ILC side. Not sure how we can implement this directly in templates.

b1ff: you just modify master templates directly on the registry

m2broth: Do you mean adding  to the templates directly? but if we do this, we will remove NewRelic not only for LDE. Could you elaborate more on your solution?

b1ff: Supposedly this tags should be implemented in a way they are removed only for LDE.
If we add these tags automatically we rely on the regexp to identify newrelic script within the template, which is a bad in a couple ways:
1) we adding new reliance to newrelic, while it is better to keep away it from ILC core
2) mechanics is not very reusable then, for other possible tools except newrelic, like sentry, datadog, etc. and not only limited to monitoring
3) newrelic might change the script making this regexp outdates, and functionality very fragile

m2broth: Do I understand right your solution:
1. Update template directly in the repository by wrapping NewRelic scripts in some custom tag
2. Add logic on ILC that removes all parts from the template that are wrapped in a custom tag (etc. if request is in LDE mode)

If it is true, how correctly update templates (by running some migration script or update them manually)? Do I understand right that only **master** template should be updated?

b1ff: Yeap, solution sound correct. 
As for changing templates there is no automated migration, as for our case we have around 4-6 master templates, it shouldn't be a big effort to update them.

ilc/server/tailor/configs-injector.js
```diff
@@ -11,6 +11,7 @@ module.exports = class ConfigsInjector {
     #cdnUrl;
     #jsInjectionPlaceholder = '<!-- ILC_JS -->';
     #cssInjectionPlaceholder = '<!-- ILC_CSS -->';
+    #markedProdTags = /<!-- Prod only start -->.*?<!-- Prod only end -->/gims;
```
blackrabbit99, line 14: what is gims ?

blackrabbit99: does not it possible to handle on tailor-x level ?

m2broth: As we discussed with @b1ff - it is better to do LDE part removal on the ILC side

m2broth: > what is gims ?

This is for capturing correctly marked tags: include across multiple lines (s flag), The g flag ensures that it finds all occurrences in the input, the i flag ensures that it is case-insensitive, and the m flag ensures that the regular expression respects line breaks when matching the start or end of lines.

wRLSS (approved):

blackrabbit99 (approved):

PR #619, opened 2024-10-10
fix(ilc): fix removing css link for embedded app

m2broth:
Fixed removing embedded application CSS link when application is unmounted during route change

### DIFF
--- a/ilc/client/ClientRouter.js
+++ b/ilc/client/ClientRouter.js
@@ -35,8 +35,6 @@ export default class ClientRouter extends EventEmitter {
         current: {},
     };
 
-    static isRouteChangeStarted = false;
-
     constructor(
         ilcConfigRoot,
         state,
@@ -276,11 +274,9 @@ export default class ClientRouter extends EventEmitter {
                     if (updateEventsToTrigger.length) {
                         updateEventsToTrigger.forEach(this.emit.bind(this));
                     }
-                    ClientRouter.isRouteChangeStarted = false;
                 },
                 { once: true },
             );
-            ClientRouter.isRouteChangeStarted = true;
         }
     };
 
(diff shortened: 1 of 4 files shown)


ilc/client/ClientRouter.js
```diff
@@ -274,9 +276,11 @@ export default class ClientRouter extends EventEmitter {
                     if (updateEventsToTrigger.length) {
                         updateEventsToTrigger.forEach(this.emit.bind(this));
                     }
+                    ClientRouter.isRouteChangeStarted = false;
```
wRLSS, line 279: dont we have some more reliant way, e.g subscribing to internal events? 

also I have small assumption on if we trigger this change multiple times it might lead to inconsistent Router state since we have it as static, @stas-nc @yehor-manzhula @b1ff please validate this assumption

m2broth: I was thinking about subscribing to internal events in CSS tracked app but this method will create additional event listeners on page (as many as we have applications on page). The isRouteChangeStarted static variable shouldn't be inconsistent because we rely on **before router change** and **after route change**  events

wRLSS (approved):

ilc/client/CssTrackedApps.spec.js
```diff
@@ -206,4 +207,51 @@ describe('CssTrackedApp', function () {
...
+            sinon.restore();
+        });
+
+        it('should mark link for removal if route change action is in progress', async () => {
+            await newApp.mount();
+
+            let link = document.querySelector(`link[href="${cssLink}"]`);
+            expect(link.getAttribute(CssTrackedApp.linkUsagesAttribute)).to.equal('1');
+
+            // Dispatch the BEFORE_ROUTING event
```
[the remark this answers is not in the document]
m2broth: yes, odd one. removed

yehor-manzhula (approved):

PR #620, opened 2024-10-16
chore: add comments about new updates to CssTrackedApp

m2broth:


ilc/client/CssTrackedApp.js
```diff
@@ -22,8 +23,10 @@ export class CssTrackedApp {
         this.#cssLinkUri = cssLink;
         this.#delayCssRemoval = delayCssRemoval;
 
-        // add route change listener for embedded apps
         if (!delayCssRemoval) {
+            // In an embedded app, removing CSS immediately after unmounting can cause visual glitches
+            // while a spinner is loading and the route is changing. Delaying CSS removal until the route
```
b1ff, line 28: I'd better to write something like this

```
While CSS for an application rendered by another application is not always immediately necessary upon unmount, there is a non-trivial case to consider:
- When the route changes and a spinner is enabled in the registry, the root application is unmounted and destroyed.
- ILC then shows a copy of the previously rendered DOM node.
- Which leads to a situation where both the root and inner applications unmount synchronously. Despite being unmounted, their styles are still required until the route transition is complete.
```

b1ff (approved):

PR #623, opened 2024-10-23
fix: redirect to 301 instead of 302 in case when only trailing slashes are present

wRLSS:


ilc/server/i18n.spec.js
```diff
@@ -201,6 +201,27 @@ describe('i18n', () => {
...
 
+            it('redirects trailing slash with 301 code', async () => {
+                const detectedI18nConfig = {
+                    locale: 'en-US',
+                    currency: 'USD',
+                };
+
+                i18nParamsDetectionPlugin.detectI18nConfig.onFirstCall().returns(detectedI18nConfig);
+
+                const req = getReqMock('///');
```
b1ff, line 212: case should be a little bit different

`/domains` -> 301 -> `/domains/`

b1ff (changes requested):

ilc/server/i18n.spec.js
b1ff: seems this is more appropriate place to make a change

https://github.com/namecheap/ilc/blob/320723cad4addeaa49553dc52c52035aa6e37c9b/ilc/server/routes/wildcardRequestHandlerFactory.ts#L47

wRLSS: Oh, I see now, I misinterpret it based on our chat conversation. Do you think the current case provides any value and we should keep 301 here also?

b1ff: To be honest unsure, while `///` is not expected and keeping normalized connection in the browser cache is more or less good, but at the same time it is unlikely that users will have that on the site, while different security scanners and attacks does not care too much whether it is 301 or 302, moreover if accidently we introduce (or it was there) a vulnerability, that depends on this case then the result will left cached in user browser and we will have limited amount of ways to fix it.

So subjectively I'd better not to push it to main rather than push

b1ff (approved):

PR #624, opened 2024-10-29
fix(ilc): fix unmount css for embedded application by adding timeout

m2broth:
This PR addresses an issue with CSS removal timing during component unmounting, particularly when components are wrapped for smooth transitions. The key improvement is the use of the setTimeout function to delay CSS removal. This ensures that the CSS is only removed after the component has been fully unmounted, preventing the visual glitch where the component briefly appears without styles.

**Key Changes:**
We now use setTimeout to delay CSS removal, ensuring that the component is fully unmounted before its styles are removed. This avoids cases where CSS is removed too early, causing visual issues.
There are scenarios where the embedded app is wrapped by another component that clones its children for smoother transitions. In such cases, removing CSS for the original node could break the styles for the cloned node. This update addresses that by allowing a delay for CSS removal, ensuring that both the original and cloned nodes retain their styles until they are fully unmounted.

**Additional Refactors:**
TypeScript Migration: The CssTrackedApp class was refactored from JavaScript to TypeScript
Refactored Options Handling: A new options property was added to CssTrackedApp, allowing more flexibility. The delayCssRemoval flag is now handled through this options object.
New removeCssTimeout Option: Introduced the removeCssTimeout option, which can be set when calling createNew({ appConfig: { removeCssTimeout: ... } }). This allows users to specify a custom delay for CSS removal after the component unmounts.


ilc/client/types/ILCAdapter.ts
stas-nc, line 1: I think this interface refers to the internal library. Should it be here?

m2broth: @stas-nc Thank you for suggestion! Totally agree with using shared interface. Going to fix this

stas-nc (approved):

PR #626, opened 2024-11-12
perf: fix cpu usage in cache

stas-nc:


stas-nc:
Took flame chart during a load test to the local ILC server connected to the production registry (see attached file) in the jira ticket.
It showed a few hot functions, one of the most important is CacheWrapper.getCache
er getCache, which under the hood makes JSON.parse on each call (because storage internally uses local storage polyfill).
I have rewritten that part to use a pure in-memory cache on the server side and native local storage on the client side.
There are also a few more hot functions that will be handled in tailor lib and maybe here in separate PRs.

ilc/common/MemoryCacheStorage.ts
```diff
@@ -0,0 +1,12 @@
+import { CacheResult, CacheStorage } from './types/CacheWrapper';
+
+export class MemoryCacheStorage implements CacheStorage {
```
wRLSS, line 3: Since we do not limit cache in any way it might lead to memory leak during long-running processes. I suggest create that instead we create `EvictingCacheStorage` based on some sort of LRU algorithm

yehor-manzhula: @wRLSS Despite the fact that your comment is 100% valid, I'd suggest addressing it within the scope of another task because, as far as I can see, the approach to storing items in memory hasn't changed. This PR simply removes serialization/deserialization, so it won't make things any worse.

wRLSS: I think the change is not that big and since we are fine-tuning ILC it can be done, small class update based on gpt suggestion without large impact. But I'm fine doing this in other PR, up to @stas-nc

stas-nc: I can do it, but was not sure, because we have some kind of pre-heat logic, in this case ig server is idle we are losing it's benefit.

wRLSS: > I can do it, but was not sure, because we have some kind of pre-heat logic, in this case, if the ilc server is idle we are losing it's benefit.

It is possible to do evicting focused on number of usage rather than time so idleness won't affect it. Anyways, approving this PR, but I think we should check this case

wRLSS (approved):

blackrabbit99 (approved):

b1ff (approved):

PR #642, opened 2025-04-02
fix(ilc server): override domain specific route only on same domain

stas-nc:


b1ff (approved):

ilc/server/tailor/merge-configs.spec.ts
```diff
@@ -390,14 +417,97 @@ describe('merge configs', () => {
...
+                        orderPos: 999,
+                        slots: {},
+                        meta: {}, // Added meta
+                        versionId: 'v1.0.5', // Added versionId
+                        domainId: 7,
+                    },
+                ],
+            };
+
+            const mergedConfig = {
```
wRLSS, line 481: Can we not perform any actual logic that can change the result of merging? I think merging function should check explicitly objects that we are awaiting. Destructurization here is a part of the logic itself which may lead to not something we are expecting. 

Let's do dumb check here

blackrabbit99 (approved):

wRLSS (approved):

PR #652, opened 2025-05-06
chore: bump webpack version

m2broth:
This PR updates the Webpack version to ensure compatibility with the latest @types/node. The recent update to @types/node introduced type changes that caused build failures in the ILC pipeline. Upgrading Webpack resolves these issues and restores a successful build.

Changes:
Updated Webpack to a version compatible with the new @types/node


m2broth:
> Should we fix webpack ilc registry as well?

yes, good point

m2broth:
> Should we fix webpack ilc registry as well?

I will update images both in ilcserver and ilcregistry

PR #699, opened 2026-06-11
feat: server-side A/B experiment assignment (Phase 1)

wRLSS:
Introduces server-side A/B experiment assignment in ILC (Phase 1).

ILC resolves a single experiment variant per visitor in the `onRequest` hook, before any fragment renders, and propagates it to every fragment via `appProps.experiments` and to the browser via inlined `ilcState` + sticky cookies. Assignment is deterministic (hash-based), cookie-sticky, config-driven, and fail-safe — any failure falls back to the baseline variant. A global kill-switch turns the whole layer off without a deploy.

See `docs/ab-testing.md` for the full design: what Phase 1 supports (and what it intentionally does not do yet), how to define experiments in configuration, and how apps read the assigned variant.

The consuming side — a `useExperiment` hook plus generic `appProps` delivery — ships as a companion change in the React application library.

### DIFF
--- a/ilc/client/ClientRouter.js
+++ b/ilc/client/ClientRouter.js
@@ -20,6 +20,7 @@ export default class ClientRouter extends EventEmitter {
     #logger;
     #registryConf;
     #ilcConfigRoot;
+    #ilcState;
     /** @type Object<Router> */
     #router;
     #prevRoute;
@@ -54,6 +55,7 @@ export default class ClientRouter extends EventEmitter {
         this.#location = location;
         this.#logger = logger;
         this.#i18n = i18n;
+        this.#ilcState = state ?? {};
         this.#ilcConfigRoot = ilcConfigRoot;
         this.#registryConf = this.#ilcConfigRoot.getConfig();
         this.#router = new Router(this.#registryConf);
@@ -142,12 +144,27 @@ export default class ClientRouter extends EventEmitter {
         if (routeConfig === undefined) {
             throw new this.errors.RouterError({ message: 'Can not find info about the slot.', data: { slotName } });
         }
-        const appProps = appConfig.props || {};
+        // ILC-level props for the app/slot. `appConfig.props` is the app's registry
+        // config; the user-app props live under a nested `appProps` key (below).
+        const appConfigProps = appConfig.props || {};
         const routeProps = routeConfig.props || {};
 
-        const finalRouteProps = deepmerge(appProps, routeProps);
+        // Carry the experiment assignments (resolved once at SSR and inlined into
+        // ilcState) into every app's props on client-side navigation. They go under the
+        // nested `appProps` key — the same place the server uses (server-router.js) and
+        // the field a client consumer reads via getCurrentPathProps().appProps —
+        // so a freshly client-mounted app still receives the already-resolved assignment.
+        // This reuses the existing value (never re-buckets), keeping the assignment stable
+        // across all apps in a long-lived SPA session.
+        //
+        // Note: SSR-rendered (primary) apps already get experiments via the server SSR
+        // appProps, so this client-side merge is the forward-looking half of that same
+        // contract — it is what delivers the assignment to apps mounted purely on the
+        // client (no SSR). It is a no-op when ilcState carries no experiments.
+        const { experiments } = this.#ilcState;
+        const experimentsProps = experiments ? { appProps: { experiments } } : {};
 
-        return finalRouteProps;
+        return deepmerge.all([appConfigProps, routeProps, experimentsProps]);
     }
 
     #setInitialRoutes = (state) => {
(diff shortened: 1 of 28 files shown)


ilc/server/experiments/index.ts
```diff
@@ -0,0 +1,107 @@
...
+        return;
+    }
+
+    request.raw.ilcState = request.raw.ilcState ?? {};
+
+    // Fully inert when disabled: leave `experiments` unset (rather than an empty object)
+    // so neither server-router nor ClientRouter merges an empty `experiments` into every
+    // app's appProps, and nothing extra is inlined into the page.
+    if (!experimentsEnabled()) {
+        return;
```
m2broth: Shouldn't we expire stale cookies when experiments are disabled? eg. users who already have x-ab-* cookies keep readable variant cookies for up to 90 days. Clients can continue seeing/applying a variant even while the kill-switch is expected to put everyone in control.

ilc/config/custom-environment-variables.json5
```diff
@@ -28,5 +28,8 @@
     },
     client: {
         protocol: 'ILC_CLIENT_PROTOCOL',
+    },
+    experiments: {
+        enabled: 'EXPERIMENTS_ENABLED',
```
stas-nc: prefix with ILC_? Not sure, but it looks like half of the variables follow this convention.

ilc/client/ClientRouter.js
```diff
@@ -142,12 +144,27 @@ export default class ClientRouter extends EventEmitter {
...
+        // across all apps in a long-lived SPA session.
+        //
+        // Note: SSR-rendered (primary) apps already get experiments via the server SSR
+        // appProps, so this client-side merge is the forward-looking half of that same
+        // contract — it is what delivers the assignment to apps mounted purely on the
+        // client (no SSR). It is a no-op when ilcState carries no experiments.
+        const { experiments } = this.#ilcState;
+        const experimentsProps = experiments ? { appProps: { experiments } } : {};
+
+        return deepmerge.all([appConfigProps, routeProps, experimentsProps]);
```
blackrabbit99, line 167: a bit expensive operation. I do not see alternative now but it could be a bottleneck potentially

ilc/server/experiments/assign.ts
```diff
@@ -0,0 +1,134 @@
...
+ * (which are later inlined into the page), and a variant removed from the ruleset
+ * is re-resolved rather than trusted.
+ *
+ * Any `x-ab-*` cookie that no longer maps to a live assignment (experiment paused
+ * or removed from the ruleset) is expired, so stale assignments don't linger for
+ * up to 90 days or resurrect if an id is reused.
+ *
+ * @param options.secure emit cookies with `Secure` (set when the site is https).
+ */
+export function assignExperiments(
```
blackrabbit99, line 50: Do you use function by purpose ? I could be complex to scale functions maybe it make sense to create responsible class ?

ilc/server/experiments/bucket.ts
```diff
@@ -0,0 +1,53 @@
+import { createHash } from 'node:crypto';
+import type { ExperimentId, ExperimentVariant, VariantName } from './interfaces';
+
+const BUCKET_COUNT = 100;
```
blackrabbit99, line 4: Does it make sense to put it into configuration ?

ilc/server/experiments/cookies.ts
```diff
@@ -0,0 +1,42 @@
+import type { CookieOptions } from './interfaces';
+
+// ILC mints its own session id (gateway-agnostic, ILC-public like `ilc-i18n`), so the
+// feature works in OSS ILC without depending on any deployment's identity headers.
+/** Stable per-visitor session id minted by ILC; seeds deterministic bucketing. */
+export const SESSION_COOKIE = 'ilc-sid';
```
blackrabbit99, line 10: point to back here.

blackrabbit99: Who is responsible for establish the cookie ?

ilc/server/experiments/bucket.ts
wRLSS: I don't think to be honest at the moment it's valuable enough, it's actually should be considered in percentage mass - that's why it's like not any number is fit here

m2broth (approved):

ilc/server/experiments/assign.ts
wRLSS: since it's a pure resolver without any internal mutable state to encapsulate and safe to call once per request with no lifecycle. constructor would add additional step without owning any state

ilc/server/experiments/cookies.ts
wRLSS: added additional comments regarding this

